PR #5, opened 2016-03-04
MET-872: Add delete document API to restlastic search client

CCheSumo:
Added delete document by query api
- Added unit test
- Run unit test

### DIFF
--- a/elasticsearch-core/src/test/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClientTest.scala
+++ b/elasticsearch-core/src/test/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClientTest.scala
@@ -146,6 +146,25 @@ class RestlasticSearchClientTest extends WordSpec with Matchers with ScalaFuture
         res.jsonStr should include(IndexName)
       }
     }
+
+    "Support delete documents" in {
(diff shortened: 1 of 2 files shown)


williamleferrand:
lgtm, thanks!

elasticsearch-core/src/test/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClientTest.scala
```diff
@@ -146,6 +146,25 @@ class RestlasticSearchClientTest extends WordSpec with Matchers with ScalaFuture
...
+      whenReady(ir) { ir =>
+        ir.created should be(true)
+      }
+      refresh()
+      val resFut = restClient.query(index, tpe, QueryRoot(TermQuery("text7", "here7")))
+      whenReady(resFut) { res =>
+        res.sourceAsMap.toList should be(List(Map("text7" -> "here7")))
+      }
+      val delFut = restClient.deleteDocument(index, tpe, QueryRoot(TermQuery("text7", "here7")))
+      val res = Await.result(delFut, 10.seconds)
```
rcoh, line 161: probably better to use `whenReady` just so we're consistent

PR #6, opened 2016-03-05
adding more query parameters

williamleferrand:


rcoh:
This change accidentally removes the snapshot version which will cause
issues in future releases and local builds.

PR #7, opened 2016-03-07
Incorporate Russell's feedbacks and fix version (1.0.3 -> 1.0.3-SNAPS…

CCheSumo:
- Fixed version from "1.0.3" to "1.0.3-SNAPSHOT"

Will update the test to WhenReady after updating to return type of deleteIndex and deleteDocument. Currently they are RawJsonResponse and cannot be easily combined with WhenReady.


CCheSumo:
plox test

PR #9, opened 2016-03-14
Add TimeStampMapping to MappingDsl

CCheSumo:
Added TimeStampMapping to Mapping Dsl

TimeStampMapping is of the following format

```
      "_timestamp": { 
        "enabled": true
      }
```

https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-timestamp-field.html


elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/dsl/MappingDsl.scala
```diff
@@ -94,6 +95,11 @@ trait MappingDsl extends DslCommons {
     }
   }
 
+  case class EnabledFiledMapping(enabled: Boolean) extends FieldMapping {
```
rcoh, line 98: I think you mean `EnabledFieldMapping`

PR #13, opened 2016-04-04
limit aws dependencies to aws-java-sdk-core

ogrodnek:
Hi!  Thanks for a great library!

I am using this in aws lambda, and the code deploy size makes a big difference there.

We only need aws-java-sdk-core for the signing utilities, so we can drop the rest of the aws package.

thanks!


weirded:
LGTM, thanks for the contribution!

PR #14, opened 2016-04-12
MET-1334: Tiny change to add capability of making a filed to be not indexed

CCheSumo:
Tiny change to add capability of making a filed to be not indexed. Also updated version numbers.

> The index attribute controls how the string will be indexed. It can contain one of three values:
> analyzed
> not_analyzed
> no

### DIFF
--- a/elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/dsl/MappingDsl.scala
+++ b/elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/dsl/MappingDsl.scala
@@ -39,9 +39,13 @@ trait MappingDsl extends DslCommons {
   sealed trait IndexType {
     val rep: String
   }
-
+  
   case object NotAnalyzedIndex extends IndexType {
     val rep = "not_analyzed"
+  }
+
+  case object NotIndexedIndex extends IndexType {
+    val rep = "no"
   }
 
   case object MappingPath {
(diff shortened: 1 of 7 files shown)


elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/dsl/MappingDsl.scala
```diff
@@ -39,11 +39,15 @@ trait MappingDsl extends DslCommons {
   sealed trait IndexType {
     val rep: String
   }
-
+  
   case object NotAnalyzedIndex extends IndexType {
     val rep = "not_analyzed"
   }
 
+  case object NotIndexedIndex extends IndexType {
```
rcoh, line 47: A comment explaining why someone might want this seemingly nonsensical mode or a link to the docs could be useful.

CCheSumo, line 47: We would like to maintain certain fields unindexed such that it is available when we retrieve the document, but do not take space and cpu when indexing.

davidandrzej, line 47: Could have some in-code commenting explaining what these mean and why you'd use them.

rcoh, line 47: Yeah what @davidandrzej said was what I meant ;-)

PR #15, opened 2016-04-18
Added Range query to QueryDsl.

ben-barnes:
Added a simple range query to the query DSL. `Range` takes a key and a list of bounds, each of which may be one of `Lt`, `Lte`, `Gt`, `Gte`.

No validation is performed; multiple of the same bound may be present. If no bounds are provided, an empty JSON object is generated, which was ignored by Elasticsearch 1.5.2.

Tested several generated JSON queries against my ElasticSearch cluster, worked fine.

### DIFF
--- a/elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/dsl/QueryDsl.scala
+++ b/elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/dsl/QueryDsl.scala
@@ -87,6 +87,39 @@ trait QueryDsl extends DslCommons {
     }
   }
 
+  case class Range(key: String, bounds: RangeBound*) extends Query {
+    val _range = "range"
+    val boundsMap = Map(key -> (bounds :\ Map[String, Any]())(_.toJson ++ _))
+
+    override def toJson: Map[String, Any] =  Map(_range -> boundsMap)
+  }
+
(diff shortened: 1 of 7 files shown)


elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/dsl/QueryDsl.scala
```diff
@@ -87,6 +87,39 @@ trait QueryDsl extends DslCommons {
     }
   }
 
+  case class Range(key: String, bounds: RangeBound*) extends Query {
```
CCheSumo: Consider using scala List to replace RangeBound*

CCheSumo:
Please add unit tests in RestlasticSearchClientTest

davidandrzej:
Thanks for the PR! Please let us know if it isn't clear how to extend `RestlasticSearchClientTest` in case we could use better commenting / documentation around that.

ben-barnes:
Sorry about missing the tests! Once I had a look at the source for `ElasticsearchIntegrationTest` it was pretty clear what was going on, and I was able to base the new tests off existing ones.

I placed several assertions in the one unit test to avoid duplicating the document insertion code. Not the best practice, but I think it fits in with the tests that are already there. Please let me know if you'd prefer that I split them up.

CCheSumo:
Merged

PR #16, opened 2016-04-19
Update autocomplete analyzer from keyword to simple to make autocompl…

CCheSumo:
…ete case insensitive


elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/dsl/MappingDsl.scala
```diff
@@ -80,8 +80,9 @@ trait MappingDsl extends DslCommons {
 
   case class CompletionMapping(context: Map[String, CompletionContext]) extends FieldMapping {
     val _type = "type" -> "completion"
-    val _analzyer = "analyzer" -> "keyword"
-    val _sanalyzer = "search_analyzer" -> "keyword"
+    // simple analyzer does case insensitive autocomplete
+    val _analzyer = "analyzer" -> "simple"
```
rcoh, line 84: when you make this change it should be as an option, not as a forced change. You are correct to bump to 1.1.0 assuming this change is backwards incompatible.

PR #17, opened 2016-04-19
Update autocomplete analyzer to simple to make it case insensitive

CCheSumo:
[x] Changed analyzer from keyword to simple
[x] Added unit tests

### DIFF
--- a/elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/dsl/MappingDsl.scala
+++ b/elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/dsl/MappingDsl.scala
@@ -78,11 +78,12 @@ trait MappingDsl extends DslCommons {
     override def toJson: Map[String, Any] = Map(_properties -> fields.mapValues(_.toJson))
   }
 
-  case class CompletionMapping(context: Map[String, CompletionContext]) extends FieldMapping {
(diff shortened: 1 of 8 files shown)


elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/dsl/MappingDsl.scala
```diff
@@ -80,8 +80,9 @@ trait MappingDsl extends DslCommons {
 
   case class CompletionMapping(context: Map[String, CompletionContext]) extends FieldMapping {
     val _type = "type" -> "completion"
-    val _analzyer = "analyzer" -> "keyword"
-    val _sanalyzer = "search_analyzer" -> "keyword"
+    // simple analyzer does case insensitive autocomplete
+    val _analzyer = "analyzer" -> "simple"
```
rcoh: this should probably be an option rather than a forced change. Best to me seems to be a case insensitive flag to `CompletionMapping`. It should default to false so that we maintain backwards compatibility.

CCheSumo: Updated CompletionMapping to use "keyword" as default and "simple" when caseSensitive is false

PR #23, opened 2016-04-26
Use AWS session credentials if provided.

ben-barnes:
Modified `AwsRequestSigner` such that if an [`AWSSessionCredentials`](http://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/auth/AWSSessionCredentials.html) object is passed into the constructor as `awsCredentials`, `withAuthHeader` will use `AwsSessionCredentials::getSessionToken()` to populate the header `X-Amz-Security-Token`.

This allowed an instance of `elasticsearch-client` running in AWS Lambda to communicate with an AWS Elasticsearch cluster.


CCheSumo:
LGTM

PR #26, opened 2016-05-06
Add support for multi-term filtered query

CCheSumo:
[x] Added support for multi-term queried filter
 [x] Added unit tests


williamleferrand:
lgtm

PR #30, opened 2016-05-20
Make execution context an argument for ScrollClient methods that crea…

kumar-avijit:
…te a Future
This PR adds a new implicit argument to supply an `ExecutionContext` to all those `RestlasticSearchClient` methods which create a `Future`.

### DIFF
--- a/elasticsearch-akka/src/test/scala/com/sumologic/elasticsearch/akkahelpers/BulkIndexerActorTest.scala
+++ b/elasticsearch-akka/src/test/scala/com/sumologic/elasticsearch/akkahelpers/BulkIndexerActorTest.scala
@@ -60,11 +60,11 @@ with BeforeAndAfterAll with BeforeAndAfterEach with MockitoSugar with ImplicitSe
   "BulkIndexerActor" should {
     "flush every message when set to 1" in {
       maxMessages = 1
-      when(mockEs.bulkIndex(any())).thenReturn(Future.successful(Seq(BulkItem("index","type", "_id", 201, None))))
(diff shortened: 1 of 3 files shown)


williamleferrand:
lgtm!

kumar-avijit:
@rcoh I would like to get your opinion on bumping up version number for this.

This is a backward incompatible change because it adds an extra parameter to various methods exposed by `ScrollClient` and `RestlasticSearchClient`. The options I see here are:
1. Bump up major version number. New version number would be `2.0.0`.
2. Bump up minor version number. New version number would be `1.1.0`.
3. (suggested as an option by @cddude229) Restore original API + add a variant for each method that takes in execution context. Bump up version number to `1.1.0`.

I am equally fine with Option 1 and 3, which are both complaint with Semantic Versioning. Option 2 isn't. 

#### Why not just add default value to the new parameter to restore backward compatibility?

It is possible to add defaults for all current methods except for one method in `RestlasticSearchClient`. The `bulkIndex` method there is already overloaded and does not admit default parameter values.

rcoh:
another option would be to add a default parameter to the `RestlasticSearchClient` itself that provides the execution context

kumar-avijit:
Ah yes, that's another option but it unfortunately falls short for our requirements atm, which is to be able to use different thread pools for indexing and searching.

kumar-avijit:
I suppose, in that case, we could add 2 default parameters to `RestlasticSearchClient` to accommodate differentiating search thread pool from indexing thread pool.

PR #32, opened 2016-06-22
Ignore more IntelliJ files

cddude229:
I had a bunch of IML files popping up, potentially because of a
different setup.  This will ignore them.


rcoh:
LGTM

cddude229:
Going to merge despite the lower coverage; I don't think a miss on .gitignore is bad.

PR #33, opened 2016-06-22
Extend feature set

cddude229:
Bunch of small changes I needed in my work.  I think they're all backwards compatible as well.

### DIFF
--- a/elasticsearch-core/src/test/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClientTest.scala
+++ b/elasticsearch-core/src/test/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClientTest.scala
@@ -61,6 +61,17 @@ class RestlasticSearchClientTest extends WordSpec with Matchers with ScalaFuture
       whenReady(mappingFut) { _ => refresh() }
     }
 
+    "Be able to setup document mapping with ignoreAbove" in {
+      val basicFiledMapping = BasicFieldMapping(StringType, None, Some(analyzerName), ignoreAbove = Some(10000))
+      val timestampMapping = EnabledFieldMapping(true)
+      val metadataMapping = Mapping(tpe, IndexMapping(
+        Map("name" -> basicFiledMapping, "f1" -> basicFiledMapping, "suggest" -> CompletionMapping(Map("f" -> CompletionContext("name")), analyzerName)),
+        timestampMapping))
+
+      val mappingFut = restClient.putMapping(index, tpe, metadataMapping)
+      whenReady(mappingFut) { _ => refresh() }
+    }
+
     "Be able to create an index, index a document, and search it" in {
       val ir = for {
         ir <- restClient.index(index, tpe, Document("doc1", Map("text" -> "here")))
--- a/elasticsearch-core/src/test/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClientTest.scala
+++ b/elasticsearch-core/src/test/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClientTest.scala
@@ -42,6 +42,10 @@ class RestlasticSearchClientTest extends WordSpec with Matchers with ScalaFuture
     new RestlasticSearchClient(new StaticEndpoint(new Endpoint(host, port)))
   }
 
+  override def refresh(): Unit = {
+    Await.result(restClient.refresh(index), 2.seconds)
+  }
+
   "RestlasticSearchClient" should {
     "Be able to create an index and setup index setting" in {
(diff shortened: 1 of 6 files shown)


elasticsearch-core/src/test/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClientTest.scala
```diff
@@ -42,6 +42,10 @@ class RestlasticSearchClientTest extends WordSpec with Matchers with ScalaFuture
     new RestlasticSearchClient(new StaticEndpoint(new Endpoint(host, port)))
   }
 
+  override def refresh(): Unit = {
+    Await.result(restClient.refresh(index), 2.seconds)
```
rcoh, line 46: I wonder if we should name this something slightly different to differentiate what it's doing

cddude229, line 46: As far as I could tell, `ElasticsearchIntegrationTest.refresh()` executes a refresh on all indices on the in-memory ES.  This refresh the provided index using the client.  I thought it was a good way to test client's `refresh()` since they should expose the same behavior.  What would you name it instead?

elasticsearch-core/src/test/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClientTest.scala
```diff
@@ -61,6 +65,17 @@ class RestlasticSearchClientTest extends WordSpec with Matchers with ScalaFuture
...
+      val basicFiledMapping = BasicFieldMapping(StringType, None, Some(analyzerName), ignoreAbove = Some(10000))
+      val timestampMapping = EnabledFieldMapping(true)
+      val metadataMapping = Mapping(tpe, IndexMapping(
+        Map("name" -> basicFiledMapping, "f1" -> basicFiledMapping, "suggest" -> CompletionMapping(Map("f" -> CompletionContext("name")), analyzerName)),
+        timestampMapping))
+
+      val mappingFut = restClient.putMapping(index, tpe, metadataMapping)
+      whenReady(mappingFut) { _ => refresh() }
+    }
+
```
CCheSumo, line 78: I will suggest to add unit tests for MustNot, MatchQuery, MatchQuery.

cddude229, line 78: Will do.  I didn't write them because not all the other query types has them, so I'll take this chance to add both.  (I did implicitly test them via the project I'm adding these for, as it had tests that hit all those cases.)

elasticsearch-core/src/test/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClientTest.scala
rcoh, line 46: `refreshWithClient`? I agree it is a nice test of the refresh method

elasticsearch-core/src/test/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClientTest.scala
cddude229, line 78: Tests added

elasticsearch-core/src/test/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClientTest.scala
cddude229, line 46: https://github.com/SumoLogic/elasticsearch-client/pull/33/commits/1078cfa469aa089441e0fc4b0fbfe89038cc7ad1  Does that do what you had in mind?  The method gets called less now as a side-effect, and renaming the original `refresh()` method to `refreshWithClient()` doesn't feel appropriate

elasticsearch-core/src/test/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClientTest.scala
```diff
@@ -51,7 +51,7 @@ class RestlasticSearchClientTest extends WordSpec with Matchers with ScalaFuture
       val analyzer = Analyzer(analyzerName, Keyword, Lowercase)
       val indexSetting = IndexSetting(12, 1, analyzer)
       val indexFut = restClient.createIndex(index, Some(indexSetting))
-      whenReady(indexFut) { _ => refresh() }
```
rcoh, line 50: yeah this is fine -- you could have used refreshWithClient everywhere, but I don't really care.

PR #35, opened 2016-06-24
Added 2 default parameters to RestlasticSearchClient to provide searc…

kumar-avijit:
…h and index thread-pools

This PR is an attempt at fixing backward compatibility that got broken by an earlier [commit](https://github.com/SumoLogic/elasticsearch-client/pull/30/commits/8bda63cb1886c0803296984555783f70db25138b) of mine.

Following @rcoh 's idea of adding execution context as a default parameter to `RestlasticSearchClient`, I have added 2 such parameters for execution contexts to be used for indexing and search.

This has allowed me to revert all `RestlasticClient`'s methods to either their original form (that doesn't require the additional execution context parameter), or, in case of `runRawEsRequest`, to admit a default value for that parameter.

The `ScrollClient` trait has also been reinstated to its original form.

I believe that with these changes, it should be possible to publish `elasticsearch-client` with only a patch-number bump.

Many thanks to @rcoh and @cddude229 for suggestions and help. @cddude229 Could you please review it once more.

### DIFF
--- a/elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClient.scala
+++ b/elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClient.scala
@@ -37,10 +37,8 @@ import scala.concurrent.duration._
 
 trait ScrollClient {
   import Dsl._
-  def startScrollRequest(index: Index, tpe: Type, query: QueryRoot, resultWindow: String = "1m")
-                        (implicit ec: ExecutionContext): Future[ScrollId]
-  def scroll(scrollId: ScrollId, resultWindow: String = "1m")
-            (implicit ec: ExecutionContext): Future[(ScrollId, SearchResponse)]
+  def startScrollRequest(index: Index, tpe: Type, query: QueryRoot, resultWindow: String = "1m"): Future[ScrollId]
+  def scroll(scrollId: ScrollId, resultWindow: String = "1m"): Future[(ScrollId, SearchResponse)]
 }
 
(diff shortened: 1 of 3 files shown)


elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClient.scala
```diff
@@ -183,7 +190,7 @@ class RestlasticSearchClient(endpointProvider: EndpointProvider, signer: Option[
                       endpoint: String,
                       method: HttpMethod = POST,
                       query: UriQuery = UriQuery.Empty)
-                     (implicit ec: ExecutionContext): Future[RawJsonResponse] = {
```
rcoh, line 186: I don't think anyone was using this endpoint, but FWIW, this isn't strictly backwards compatible. Having a default value for a second parameter still requires a code change, I think

kumar-avijit, line 186: So this works, e.g.:

```
scala> class A {
     |   def foo(x: Int)(implicit name: String = "foo"): Int = {
     |     println (name)
     |     x + 1
     |   }
     | }
defined class A

scala> new A().foo(0)
foo
res3: Int = 1

scala> new A().foo(0)("bla")
bla
res4: Int = 1
```

rcoh, line 186: Oh good. lgtm

cddude229:
Retroactive lgtm as well.

PR #36, opened 2016-06-29
Add refresh interval index setting and support same key multi values in completion context

CCheSumo:
- Added refresh interval index setting
- Support same key multi values in completion context

### DIFF
--- a/elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/dsl/CompletionDsl.scala
+++ b/elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/dsl/CompletionDsl.scala
@@ -20,7 +20,7 @@ package com.sumologic.elasticsearch.restlastic.dsl
 
 trait CompletionDsl extends DslCommons with QueryDsl {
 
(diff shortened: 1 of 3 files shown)


elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/dsl/CompletionDsl.scala
```diff
@@ -33,7 +33,35 @@ trait CompletionDsl extends DslCommons with QueryDsl {
...
+
+    override def toJson: Map[String, Any] = Map(
+      _field -> field,
+      _context -> context,
+      _size -> size
+    )
+  }
+
+  trait CompletionWithContext {
+    def toJson: Map[String, Any]
```
rcoh: It should be possible to share the code between these 2 classes. 

```
trait CompletionWithContext {
   val _field = "field"
  def context: Iterable[(String, String)]
  def toJson: Map[String, Any] = ...
}
```

That reduces the chance of introducing bugs by changing 1 and not the other.

CCheSumo: Very good point. Will update

PR #38, opened 2016-06-30
Add optoin of refresh interval in index settings & allow context aware autocomplete without extraction from path

CCheSumo:
- Add optoin of refresh interval in index settings 
- Allow context aware autocomplete without extraction from path

We find the extraction at ingest time is very time consuming. Without the extraction, we can have 5x ingest performance boost. Therefore, we remove the extraction and statically supply all the contexts.


CCheSumo:
plox test

PR #48, opened 2016-07-26
Add Exists query

shushu:


rcoh:
Looks good -- Would you mind also adding a test to `RestlasticSearchClientTest`? See https://github.com/SumoLogic/elasticsearch-client/pull/45/files for an example.

CCheSumo:
@shushu I will be able to merge your PR with unit test. Would you be able to add the unit test?

rcoh:
I started writing a test for this but realized the exists quiet is 2.0 only. I need to figure out the best way to support 2.9 features in a sane way.

PR #60, opened 2016-08-22
BucketAggregation with nesting buckets inside other buckets

ArnaudT:
Hi,

We need to do aggregations with [nesting buckets inside other buckets](https://www.elastic.co/guide/en/elasticsearch/guide/current/_buckets_inside_buckets.html).

This is a breaking change as the object returned now a `Bucket` instead of a `BucketAggregationResultBody`
We return a map instead of a well defined case class because the buckets name can be specified in the query.

Please, tell me what you think.

Cheers,
Arnaud.

### DIFF
--- a/elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClient.scala
+++ b/elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClient.scala
@@ -83,10 +83,17 @@ class RestlasticSearchClient(endpointProvider: EndpointProvider, signer: Option[
     }
   }
 
-  def bucketAggregation(index: Index, tpe: Type, query: AggregationQuery): Future[Bucket] = {
+  def bucketNestedAggregation(index: Index, tpe: Type, query: AggregationQuery): Future[BucketNested] = {
     implicit val ec = searchExecutionCtx
     runEsCommand(query, s"/${index.name}/${tpe.name}/_search").map { rawJson =>
-      Bucket(rawJson.mappedTo[BucketAggregationResponse].aggregations._2)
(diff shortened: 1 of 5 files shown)


rcoh:
I'm leaving on vacation for 2w but hopefully someone else can review this.
Ideally we would find a way to avoid a breaking change, probably by adding
a second API

ArnaudT:
Sure,  I can create a new method like `bucketNestedAggregation` that can return embedded buckets and leave `bucketAggregation` with only one bucket aggregation level.

CCheSumo:
@ArnaudT I agree with @rcoh. Let's have a second API. `bucketNestedAggregation` sounds good to me.

ArnaudT:
I have updated the PR. 

I created a new method `bucketNestedAggregation`
I am not sure if I really like to have the following code. It's not duplicated but really similar. 

```
   case class BucketNested(underlying: BucketNestedMap)
    type BucketNestedMap = Map[String, Any]
    type NestedAggregations = (String, BucketNestedMap)
    case class BucketNestedAggregationResponse(aggregations: NestedAggregations)
    case class BucketNestedAggregationResultBody(doc_count_error_upper_bound: Int,
                                           sum_other_doc_count: Int,
                                           buckets: List[BucketNestedMap])


    case class BucketAggregationResponse(aggregations: Aggregations)
    case class Aggregations(aggs_name: BucketAggregationResultBody )
    case class BucketAggregationResultBody(doc_count_error_upper_bound: Int,
                                           sum_other_doc_count: Int,
                                           buckets: List[Bucket])
    case class Bucket(key: String, doc_count: Int)
```

I do not think of better way to do it.
What do you think?

Arnaud.

rcoh:
Codewise, LGTM. Can you run

```
mvn license:update-file-header
```

That should fix the license on the file you added and allow the CI build to pass.
Much thanks :-)

rcoh:
LGTM

PR #62, opened 2016-08-22
Refactoring to handle sorting by distance

ArnaudT:
Hi,

We need to sort the result by [distance](https://www.elastic.co/guide/en/elasticsearch/guide/current/sorting-by-distance.html).
- update QueryRoot to get more complex sorting than a string tuple
- QueryDsl extends now SortDsl
- SortDsl has `SimpleSort`,  `GeoDistanceSort` and `asc`, `desc` order
- add GeoPointType

Please, tell me what you think.

Cheers,
Arnaud.


CCheSumo:
Looks good to me. Also backward compatible. Merge.

Issue #63, opened 2016-08-30
Expired credentials provider


ArnaudT:
Hi,

We have an issue with the AWS credentials.

We are instantiating the service like in the README.md:

```
  private val signer = new AwsRequestSigner(awsCredentials, "us-east-1", "es")
  private val endpoint = new StaticEndpoint(Endpoint(elasticSearch.getHost, elasticSearch.getPort))
  private val restClient = new RestlasticSearchClient(endpoint, Some(signer))
```

We are using the credentials provider implementation that loads credentials from the Amazon EC2 Instance Metadata Service named `InstanceProfileCredentialsProvider`.
After few hours we have the following error message:

`{"message":"The security token included in the request is expired"}`

We should refresh the temporary credentials but it seems it's not possible with the current implementation of the client.

Do you have any advices about handling expiring tokens? 

One option could be to instantiate a new instance of `RestlasticSearchClient` at each request but this will instantiate a new `ActorSystem()` each time. Which is a really heavy operation. 
Furthermore, I do not think instantiating an actor system in the client is a good thing. As the akka doc says: [An ActorSystem is a heavyweight structure that will allocate 1…N Threads, so create one per logical application.](http://doc.akka.io/docs/akka/2.4/general/actor-systems.html)
One improvement could be to remove the creation of the `ActorSystem` within this class and add it as a dependency as the `indexExecutionCtx: ExecutionContext` or `searchExecutionCtx: ExecutionContext`. So we can have only one `ActorSystem` per application.

What do you think?

Some resources:
- https://aws.amazon.com/premiumsupport/knowledge-center/security-token-expired/
- http://stackoverflow.com/questions/36732171/sqs-expiredtoken-the-security-token-included-in-the-request-is-expired-status-c

Thanks,
Arnaud.

rcoh:
Fixed.

PR #64, opened 2016-08-30
AwsRequestSigner and refresh tokens

ArnaudT:
Hi,

This PR is linked to the issue #63.

It adds a new `AwsRequestSigner` constructor to handle aws tokens that needs to be refreshed.

An other way is to pass a `AWSCredentialsProvider` instead of the `AWSCredentials` and call `getCredentials()` method in `AwsRequestSigner` class.

What about a merge with the following project https://github.com/ticofab/aws-request-signer ?

What do you think?

Thanks,
Arnaud.

### DIFF
--- a/elasticsearch-aws/src/main/scala/com/sumologic/elasticsearch/util/AwsRequestSigner.scala
+++ b/elasticsearch-aws/src/main/scala/com/sumologic/elasticsearch/util/AwsRequestSigner.scala
@@ -176,3 +176,25 @@ class AwsRequestSigner(awsCredentials: AWSCredentials, region: String, service: 
 
   private def hexOf(buf: Array[Byte]) = buf.map("%02X" format _).mkString.toLowerCase
 }
+
+/** Factory for [[com.sumologic.elasticsearch.util.AwsRequestSigner]] instances. */
+object AwsRequestSigner {
+
+  /**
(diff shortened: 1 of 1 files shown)


rcoh:
Are you sure this works? It doesn't look like the credentials could be refreshed...

I would support adding another constructor that takes a credentials provider and reloads the creds every time, I think that design is a bit cleaner. 

In the long term, I think using that project as a dependency is wise, provided it is reliable. Our implementation has definitely been a source of some bugs.

 Would you mind opening an issue and tagging it with `2.0` or `breaking change`? You can also open an issue to clean up the bucket aggregator API. When we've accumulated enough of these, we can bump the version to 2.0 and make all the breaking changes at once.

Thanks so much for your help and improvements!

ArnaudT:
The new constructor takes a function as a parameter. So each time, the `AwsRequestSigner` call the parameter `awsCredentialsFunction`, it calls the `aWSCredentialsProvider.getCredentials` that will renew the credentials if needed. (I have just updated the scaladoc to make it more clear)
So the logic is moved outside of the library.  (we are testing in staging the solution and it solved our issue with expiring tokens.)

I do not think I have the right to tag the issues or the PR. 
I can propose a new PR with the AWSCredentialsProvider as a parameter for a new constructor.

Thanks,
Arnaud.

ArnaudT:
I'll create a new PR with a AWSCredentialsProvider soon. I close this one.

PR #67, opened 2016-09-05
Update AwsRequestSigner to use AWSCredentialsProvider

ArnaudT:
Hi, 

We are using the `EC2ContainerCredentialsProviderWrapper` to loads the credentials from Amazon EC2 Instance Metadata Service. These `AWSCredentials` needs to be refreshed so it's better to use the aws provider and let the provider handling the refresh logic then using directly the `AWSCredentials` class.

This change is backward compatible as we create a `StaticCredentialsProvider` if we pass the `AWSCredentials` on the second constructor.
This PR follow PR #64.

Cheers.
Arnaud.

### DIFF
--- a/elasticsearch-aws/src/main/scala/com/sumologic/elasticsearch/util/AwsRequestSigner.scala
+++ b/elasticsearch-aws/src/main/scala/com/sumologic/elasticsearch/util/AwsRequestSigner.scala
@@ -25,19 +25,33 @@ import java.util.{Calendar, TimeZone}
 import javax.crypto.Mac
 import javax.crypto.spec.SecretKeySpec
 
-import com.amazonaws.auth.{AWSCredentials, AWSSessionCredentials}
+import com.amazonaws.auth.{AWSCredentials, AWSCredentialsProvider, AWSSessionCredentials}
+import com.amazonaws.internal.StaticCredentialsProvider
 import com.sumologic.elasticsearch.restlastic.RequestSigner
 import spray.http.HttpHeaders.{Host, RawHeader}
 import spray.http.HttpRequest
 
 /**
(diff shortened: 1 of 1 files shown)


elasticsearch-aws/src/main/scala/com/sumologic/elasticsearch/util/AwsRequestSigner.scala
```diff
@@ -25,19 +25,33 @@ import java.util.{Calendar, TimeZone}
...
  * Sign AWS requests following the instructions at http://docs.aws.amazon.com/general/latest/gr/sigv4_signing.html
- * @param awsCredentials AWS creds to sign with
+ * @param awsCredentialsProvider Interface for providing AWS credentials
  * @param region Region to connect to
  * @param service Service to connect to (eg. "es" for elasticsearch) [http://docs.aws.amazon.com/general/latest/gr/rande.html]
  */
-class AwsRequestSigner(awsCredentials: AWSCredentials, region: String, service: String) extends RequestSigner {
+class AwsRequestSigner(awsCredentialsProvider: AWSCredentialsProvider, region: String, service: String) extends RequestSigner {
   val Algorithm = "AWS4-HMAC-SHA256"
+  require(Option(awsCredentialsProvider.getCredentials).isDefined, "awsCredentialsProvider must return non null awsCredentials.")
```
rcoh: Although I can see why you did this, seems like `!= Null`  would be a bit clearer

ArnaudT: sure! done.

rcoh:
LGTM. Definitely a change for the better! Thanks as always.

Issue #68, opened 2016-09-16
Add query boost


Cowa:
Hi,
Thanks for this library, it works very well.

But there is one missing feature for my case that prevent me to fully adopt it: boosting query:
https://www.elastic.co/guide/en/elasticsearch/guide/current/_boosting_query_clauses.html

I don't know if it would be difficult to implement it.
I may open a PR if I manage to take some time!

rcoh:
Looks like it would be pretty straightforward to add. 
https://github.com/SumoLogic/elasticsearch-client/blob/master/elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/dsl/QueryDsl.scala has the current query features, and it's easy to add new ones. Let me know if you have any questions.

PR #69, opened 2016-09-16
Add match query boost

Cowa:
Issue #68

It follows this format: 

``` json
"match": {
  "content": {
    "query": "Elasticsearch",
    "boost": 3 
  }
}
```

https://www.elastic.co/guide/en/elasticsearch/guide/current/_boosting_query_clauses.html

### DIFF
--- a/elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/dsl/QueryDsl.scala
+++ b/elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/dsl/QueryDsl.scala
@@ -216,15 +216,7 @@ trait QueryDsl extends DslCommons with SortDsl {
     }
   }
 
-  case class MatchQuery(key: String, value: String) extends Query {
-    val _match = "match"
-
-    override def toJson: Map[String, Any] = {
-      Map(_match -> Map(key -> value))
-    }
-  }
-
-  case class MatchQueryWithBoost(key: String, value: String, boost: Double) extends Query {
+  case class MatchQuery(key: String, value: String, boost: Double = 1) extends Query {
     val _match = "match"
     val _query = "query"
     val _boost = "boost"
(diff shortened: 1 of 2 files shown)


rcoh:
Great! Would you mind adding a test to https://github.com/SumoLogic/elasticsearch-client/blob/master/elasticsearch-core/src/test/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClientTest.scala?

Cowa:
Hi @rcoh, I added the test

rcoh:
Looks good. One last thing: Would you mind unifying `MatchQuery` and `MatchQueryWithBoost` by adding a boost argument to match query but defaulting it to none, rather than adding a new class:

```
case class MatchQuery(key: String, value: String, boost: Option[Double]) {
   ...
}
```

Cowa:
Thanks for the feedback!

By default ElasticSearch uses a boost of `1`. So we could have `boost: Double = 1` instead of `boost: Option[Double] = None`. What do you think? It will avoid the user to wrap boosts with `Some`.

**With `Option`**

``` scala
case class MatchQuery(key: String, value: String, boost: Option[Double] = None) extends Query {
    val _match = "match"
    val _query = "query"
    val _boost = "boost"

    override def toJson: Map[String, Any] = {
      Map(_match ->
        Map(key ->
          Map(_query -> value,
              _boost -> boost.getOrElse(1))))
    }
  }
```

**With `1` as default**

``` scala
case class MatchQuery(key: String, value: String, boost: Double = 1) extends Query {
    val _match = "match"
    val _query = "query"
    val _boost = "boost"

    override def toJson: Map[String, Any] = {
      Map(_match ->
        Map(key ->
          Map(_query -> value,
              _boost -> boost)))
    }
  }
```

rcoh:
Default of 1 is much better! Thanks for the help.

Cowa:
Perfect! I pushed the change.

rcoh (approved):
LGTM. Thanks for contributing!

rcoh:
Closes #69

Cowa:
You're welcome!

One last question, does the new release will be published on the maven repo soon?
I'll need it to deploy my application on production.

rcoh:
@CCheSumo would you mind cutting a new release when you get a chance?

CCheSumo:
Sure. I will do it shortly

rcoh:
@Cowa the new version has been pushed to maven central

Cowa:
Thank you! 👍

PR #70, opened 2016-09-19
Multi match query with options

ArnaudT:
Hi,

I wanted to add some options to the `MultiMatchQuery` like `"operator":   "and"`
I didn't want to break the `MultiMatchQuery` class so I created a new one called `MultiMatchQueryWithOptions` which has a new parameter `options: Map[String, String]`

https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-multi-match-query.html

Tell me what you think!

Thanks.


rcoh:
LGTM. I added https://github.com/SumoLogic/elasticsearch-client/issues/71 to track removing the code duplication when we release the new version.

Issue #66, opened 2016-09-01
SearchResponse has empty jsonStr for queries with sourceFilter 


jonasanso:
Hi there,

I am using the elastic search client and it is quite nice, so thanks for your work.

I have a query that uses sourceFilter and to my surprise in the SearchResponse jsonStr is empty.

I found the cause 

https://github.com/SumoLogic/elasticsearch-client/blob/e41b4b247cd812ae5c7336cc2494e62f83bdfba5/elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClient.scala#L81

I can workaround the issue using the rawSearchResponse. 

I just would like to know, if there is a reason for that line to be there.

Best regards,
Jonas.

rcoh:
@CCheSumo may know. Seems like it could be a bug, we'll look into it. Thanks for the report!

jonasanso:
Thanks

CCheSumo:
Hi @jonasanso, 

Any reason your are not using the RawSearchResponse? It is the parsed out version of the raw jsonStr. We decided not to pass back the raw jsonStr when source filter is supplied (for performance benefits). Again, all the same information are available in RawSearchResponse. It is recommended to use it rather than raw jsonStr.

In the extreme scenario, if you really really want to use the raw jsonStr (with source filtering enabled), you can simply convert RawSearchResponse to a raw jsonStr on the local client side. There is no need to pass the same duplicate content over the internet twice only because they are in different formats.

jonasanso:
Hi @CChesumo,

Thanks for your answer. 

I am migrating our code base from a different ES client and I already have the parsing logic to our models implemented in spray-json.

So for me is more useful the jsonStr than the RawSearchResponse. 

You are right I can get around it and formatting RawSearchResponse to string and parsing it back.

But the question is, why SearchResponse contain the same data in two different ways, string and parsed using json4s? 

For my needs, I find it OK because I can usey parsers,  other people might prefer the already parsed response. But in the current scenario where we have both values, I find it strange that the jsonStr is missing when filtering, in fact when filtering we are reducing the size of jsonStr. 

I am totally with you that there is no need to have this two attributes in memory and for the clarity of the API there should be only one. But consider the impact of forcing one json library. 

Again, thanks for your efforts,
Jonas

CCheSumo:
@jonasanso,

[...]

PR #73, opened 2016-09-20
Add option to specify whether to return rawJsonStr along with RawJsonResponse

CCheSumo:
This PR implements the design discussed in https://github.com/SumoLogic/elasticsearch-client/issues/66

The PR adds an option to let user specify
- true: return rawJsonStr along with rawJsonResponse 
- false: only return rawJsonResponse (reson we do this is that for large documents, it has certain performance impacts when we retrieve many documents with one query)

rawJsonResponse is always returned for backward compatibility.

### DIFF
--- a/elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClient.scala
+++ b/elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClient.scala
@@ -75,13 +75,10 @@ class RestlasticSearchClient(endpointProvider: EndpointProvider, signer: Option[
   import RestlasticSearchClient.ReturnTypes._
 
   def ready: Boolean = endpointProvider.ready
-  def query(index: Index, tpe: Type, query: QueryRoot, rawJsonStr: Option[Boolean] = Some(true)): Future[SearchResponse] = {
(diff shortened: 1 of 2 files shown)


CCheSumo:
@rcoh I would like to have this change in the new branch. Could you review it at your convenience? Then I can cut the branch.

elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClient.scala
```diff
@@ -75,10 +75,13 @@ class RestlasticSearchClient(endpointProvider: EndpointProvider, signer: Option[
   import RestlasticSearchClient.ReturnTypes._
 
   def ready: Boolean = endpointProvider.ready
-  def query(index: Index, tpe: Type, query: QueryRoot): Future[SearchResponse] = {
+  def query(index: Index, tpe: Type, query: QueryRoot, rawJsonStr: Option[Boolean] = Some(true)): Future[SearchResponse] = {
```
rcoh: why not just have this default to true? Why does it need to be `Some(true)`?

CCheSumo: I don't really have a preference for one versus the other. One is shorter & cleaner, the other clearly tells me it is an option. Updated to `ture` as suggested.

rcoh:
LGTM

PR #76, opened 2016-09-27
Add field index options and query dsl for highlight

CCheSumo:
- Added capability to specify index_options in field mapping 
  https://www.elastic.co/guide/en/elasticsearch/reference/current/index-options.html
- Added query dsl for highlight
  https://www.elastic.co/guide/en/elasticsearch/reference/1.5/search-request-highlighting.html#postings-highlighter
- Added highlight results in SearchResponse

### DIFF
--- a/elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClient.scala
+++ b/elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClient.scala
@@ -75,7 +75,7 @@ class RestlasticSearchClient(endpointProvider: EndpointProvider, signer: Option[
   import RestlasticSearchClient.ReturnTypes._
 
   def ready: Boolean = endpointProvider.ready
-  def query(index: Index, tpe: Type, query: QueryRootBase, rawJsonStr: Boolean = true): Future[SearchResponse] = {
+  def query(index: Index, tpe: Type, query: QueryRootLike, rawJsonStr: Boolean = true): Future[SearchResponse] = {
     implicit val ec = searchExecutionCtx
     runEsCommand(query, s"/${index.name}/${tpe.name}/_search").map { rawJson =>
       val jsonStr = if(rawJsonStr) rawJson.jsonStr else ""
--- a/elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClient.scala
+++ b/elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClient.scala
@@ -75,7 +75,7 @@ class RestlasticSearchClient(endpointProvider: EndpointProvider, signer: Option[
   import RestlasticSearchClient.ReturnTypes._
 
   def ready: Boolean = endpointProvider.ready
-  def query(index: Index, tpe: Type, query: QueryRoot, rawJsonStr: Boolean = true): Future[SearchResponse] = {
+  def query(index: Index, tpe: Type, query: QueryRootBase, rawJsonStr: Boolean = true): Future[SearchResponse] = {
     implicit val ec = searchExecutionCtx
     runEsCommand(query, s"/${index.name}/${tpe.name}/_search").map { rawJson =>
       val jsonStr = if(rawJsonStr) rawJson.jsonStr else ""
@@ -259,6 +259,7 @@ object RestlasticSearchClient {
       }
 
       def sourceAsMap: Seq[Map[String, Any]] = rawSearchResponse.sourceAsMap
+      def highlightAsMaps: Seq[Map[String, Any]] = rawSearchResponse.highlightAsMaps
       def length = rawSearchResponse.hits.hits.length
     }
 
@@ -276,6 +277,7 @@ object RestlasticSearchClient {
       }
 
       def sourceAsMap: Seq[Map[String, Any]] = hits.hits.map(_._source.values)
+      def highlightAsMaps: Seq[Map[String, Any]] = hits.hits.flatMap(_.highlight.map(_.values))
     }
 
     case class BucketNested(underlying: BucketNestedMap)
@@ -299,7 +301,8 @@ object RestlasticSearchClient {
                                    _type: String,
                                    _id: String,
                                     _score: Option[Float],
-                                   _source: JObject)
+                                   _source: JObject,
+                                   highlight: Option[JObject] = None)
 
     case class RawJsonResponse(jsonStr: String) {
       private implicit val formats = org.json4s.DefaultFormats
--- a/elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/dsl/DslCommons.scala
+++ b/elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/dsl/DslCommons.scala
@@ -32,10 +32,6 @@ trait DslCommons {
     val rep: String
   }
 
-  trait FieldIndexOption {
-    val option: String
-  }
-
   trait HighlighterType {
     val name: String
   }
--- a/elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/dsl/DslCommons.scala
+++ b/elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/dsl/DslCommons.scala
@@ -32,10 +32,6 @@ trait DslCommons {
     val rep: String
   }
 
-  trait HighlighterType {
-    val name: String
-  }
-
   object EsOperation {
     implicit val formats = org.json4s.DefaultFormats
     def compactJson(map: Map[String, Any]) = compact(render(decompose(map)))
--- a/elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/dsl/DslCommons.scala
+++ b/elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/dsl/DslCommons.scala
@@ -32,6 +32,14 @@ trait DslCommons {
     val rep: String
   }
 
+  trait FieldIndexOption {
+    val option: String
+  }
+
+  trait HighlighterType {
+    val name: String
+  }
+
   object EsOperation {
     implicit val formats = org.json4s.DefaultFormats
     def compactJson(map: Map[String, Any]) = compact(render(decompose(map)))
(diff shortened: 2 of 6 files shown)


CCheSumo:
@rcoh Can you take a look at this code changes? It enables "highlight" in the query.

rcoh:
Will take a look this evening.

elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClient.scala
```diff
@@ -299,7 +301,8 @@ object RestlasticSearchClient {
                                    _type: String,
                                    _id: String,
                                     _score: Option[Float],
-                                   _source: JObject)
+                                   _source: JObject,
+                                   highlight: Option[JObject] = None)
```
rcoh, line 305: please file an issue to remove this default in 2.0

elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/dsl/DslCommons.scala
```diff
@@ -32,6 +32,14 @@ trait DslCommons {
     val rep: String
   }
 
+  trait FieldIndexOption {
```
rcoh: This can just go inside `MappingDsl` and then it can be sealed

elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/dsl/MappingDsl.scala
```diff
@@ -119,16 +135,19 @@ trait MappingDsl extends DslCommons {
   val _analyzer = "analyzer"
   val _searchAnalyzer = "search_analyzer"
   val _ignoreAbove = "ignore_above"
+  val _fieldIndexOpions = "index_options"
 
   case class BasicFieldMapping(tpe: FieldType, index: Option[IndexType], analyzer: Option[Name],
-                               ignoreAbove: Option[Int] = None, search_analyzer: Option[Name]= None)
+                               ignoreAbove: Option[Int] = None, search_analyzer: Option[Name]= None,
+                               fieldIndexOption: Option[FieldIndexOption] = None)
```
rcoh: why not call this `indexOption` to be consistent with the other terms?

elasticsearch-core/src/test/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClientTest.scala
```diff
@@ -86,6 +86,27 @@ class RestlasticSearchClientTest extends WordSpec with Matchers with ScalaFuture
...
+
+      val timestampMapping = EnabledFieldMapping(true)
+      val metadataMapping = Mapping(tpe, IndexMapping(
+        Map("name" -> basicFieldDocsMapping, "f1" -> basicFieldFreqsMapping,
+          "f2" -> basicFieldOffsetsMapping, "text" -> basicFieldOffsetsMapping,
+          "suggest" -> CompletionMapping(Map("f" -> CompletionContext("name")),
+            analyzerName)), timestampMapping))
+
+      val mappingFut = restClient.putMapping(index, tpe, metadataMapping)
+      whenReady(mappingFut) { _ => refresh() }
```
rcoh, line 107: This test only checks that you put the mapping and it happens to not error (if we even propagate errors properly!) Can you make a call to `getMapping` to verify that the mapping was actually applied?

elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/dsl/QueryDsl.scala
```diff
@@ -368,4 +370,64 @@ trait QueryDsl extends DslCommons with SortDsl {
       _dis_max -> innerMap
     )
   }
+
+  case class HighlightRoot(queryRoot: QueryRoot, highlight: Highlight)
```
rcoh, line 378: maybe I'm missing something -- can we just have HighlightRoot extend RootObject rather than creating a separate `QueryRootBase`?

rcoh: maybe I'm missing something -- can we just have `HighlightRoot` extend RootObject rather than creating a separate `QueryRootBase`?

elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/dsl/QueryDsl.scala
```diff
@@ -368,4 +370,64 @@ trait QueryDsl extends DslCommons with SortDsl {
...
+    val _post_tags = "post_tags"
+    val _fields = "fields"
+    val _highlight = "highlight"
+
+    val pre_tags = if (preTags.isEmpty) Map[String, Any]() else Map(_pre_tags -> preTags)
+    val post_tags = if (postTags.isEmpty) Map[String, Any]() else Map(_post_tags -> postTags)
+
+    override def toJson: Map[String, Any] = Map(
+      _highlight -> {
+        Map( _fields -> fields.foldLeft(Map[String, Any]())((l, r) => l ++ r.toJson)) ++
```
rcoh: this code is pretty hard to read. I would probably do it as:
`fields.map(_.toJson).reduce(_ ++ _)` what do you think?

rcoh (commented):
Need to take a bit more time to read all of it. Some comments so far.

elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClient.scala
CCheSumo: Issue filed as requested https://github.com/SumoLogic/elasticsearch-client/issues/77

elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/dsl/DslCommons.scala
CCheSumo: Moved to be a sealed trait in MappingDsl as requested.

elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/dsl/MappingDsl.scala
CCheSumo: Updated as requested.

elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/dsl/QueryDsl.scala
CCheSumo: Oh, making a separate `QueryRootBase` is to ensure `query` must take one of its two implementations (either `QueryRoot`, or `HighlightRoot`) and throw error at compile time if other type of parameter is provided.

CCheSumo: Not sure why this is duplicated. Please refer to the comments above.

elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/dsl/QueryDsl.scala
CCheSumo: Ah, I should have thought of using reduce instead of foldLeft. It makes the code much cleaner! Thanks for the advice. Updated.

elasticsearch-core/src/test/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClientTest.scala
CCheSumo: Added code to verify mapping using getMapping.

CCheSumo:
@rcoh Thanks for the feedbacks. I should have addressed most of the concerns raised. I will wait for your second review before merging.

elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/dsl/DslCommons.scala
```diff
@@ -32,6 +32,10 @@ trait DslCommons {
     val rep: String
   }
 
+  trait HighlighterType {
```
rcoh: seems like this could move into QueryDsl as well since it's only used there?

elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/dsl/MappingDsl.scala
```diff
@@ -93,6 +97,25 @@ trait MappingDsl extends DslCommons {
     }
   }
 
+  // Supported in elasticsearch v2.4
```
rcoh, line 100: I think we need a formalized way of handling elasticsearch 2.x stuff. Ideally it would be a different DSL you could import or something. Another task to think about pre-2.0.

elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/dsl/QueryDsl.scala
rcoh: I see -- you don't want a HighlightRoot to be nested inside of a HighlightRoot

rcoh: Can you rename it to `QueryRootLike`? That's the scala standard way for saying something is a trait.

rcoh (approved):
LGTM -- see a couple more small comments above.

elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/dsl/DslCommons.scala
CCheSumo: Moved to QueryDsl as a sealed trait

elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/dsl/QueryDsl.scala
CCheSumo: Updated QueryRootBase to QueryRootLike

PR #79, opened 2016-09-30
Add option of terminate_after when run a query

CCheSumo:
It turns out terminate_after can have huge impact on latency when the document counts are large. 

In order to support terminate_after and maintain backward compatibility, I have to create an ExtendedQueryRoot that inherits QueryRoot. This api can serve to extend any further enhancements to QueryRoot before 2.0. 

This api should be incorporated into QueryRoot in 2.0 when breaking changes are allowed.

### DIFF
--- a/elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/dsl/QueryDsl.scala
+++ b/elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/dsl/QueryDsl.scala
@@ -70,11 +70,9 @@ trait QueryDsl extends DslCommons with SortDsl {
       val sorts = sortOpt
         .map(_.foldLeft(Seq.empty[Sort])((sorts, value) => sorts :+ SimpleSort(value._1, SortOrder.fromString(value._2))))
         .getOrElse(Nil)
-      if (terminateAfter.isDefined) {
-        new ExtendedQueryRoot(query, fromOpt, sizeOpt, sorts, timeout, sourceFilter, terminateAfter)
-      } else {
-        new QueryRoot(query, fromOpt, sizeOpt, sorts, timeout, sourceFilter) 
-      }
+      terminateAfter
+        .map(tA => new ExtendedQueryRoot(query, fromOpt, sizeOpt, sorts, timeout, sourceFilter, Some(tA)))
+        .getOrElse(new QueryRoot(query, fromOpt, sizeOpt, sorts, timeout, sourceFilter))
     }
   }
   
--- a/elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/dsl/QueryDsl.scala
+++ b/elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/dsl/QueryDsl.scala
@@ -72,6 +72,24 @@ trait QueryDsl extends DslCommons with SortDsl {
       new QueryRoot(query, fromOpt, sizeOpt, sorts, timeout, sourceFilter)
     }
   }
+  
+  // This api should be incorporated into QueryRoot in 2.0 when breaking changes are allowed.
+  // Before that, this api can serve to extend QueryRoot with backward compatibility.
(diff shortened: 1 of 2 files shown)


CCheSumo:
@rcoh PR for enabling terminate_after when running a query.

elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/dsl/QueryDsl.scala
```diff
@@ -72,6 +72,24 @@ trait QueryDsl extends DslCommons with SortDsl {
       new QueryRoot(query, fromOpt, sizeOpt, sorts, timeout, sourceFilter)
     }
   }
+  
+  // This api should be incorporated into QueryRoot in 2.0 when breaking changes are allowed.
+  // Before that, this api can serve to extend QueryRoot with backward compatibility.
+  class ExtendedQueryRoot(override val query: Query,
```
rcoh, line 81: I must be missing something. Why can't we just add terminateAfter with a default value and keep backwards compat?

CCheSumo: We have `QueryRoot` object which has an `apply` method that has parameters with default values. `QueryRoot` case class also has an `apply` method. 

Currently, all `QueryRoot` case class's  parameters do not have default values, which is fine. 

However, if we were to add `terminateAfter: Option[Int]=None`, the code will not compile due to scala compiler disallowing overloaded methods with default arguments.

CCheSumo: Related read https://groups.google.com/forum/#!msg/scala-user/FyQK3-cqfaY/fXLHr8QsW_0J

rcoh:
OK LGTM. I made a suggestion above you can take if you want.

Seems like we're building up enough stuff to cut 2.0 soon.

elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/dsl/QueryDsl.scala
rcoh: oh right! Forgot about that bit of nastiness. If you want, you could add another optional parameter to the apply method. If `terminateAfter` was defined, we'd return an `ExtendedQueryRoot` instead. That way we hide the problem inside of the library. What do you think?

CCheSumo: That makes sense. Let me put in that.

elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/dsl/QueryDsl.scala
```diff
@@ -65,11 +65,16 @@ trait QueryDsl extends DslCommons with SortDsl {
...
               sortOpt: Option[Seq[(String, String)]] = None,
               timeout: Option[Int] = None,
-              sourceFilter: Option[Seq[String]] = None): QueryRoot = {
+              sourceFilter: Option[Seq[String]] = None,
+              terminateAfter: Option[Int] = None): QueryRoot = {
       val sorts = sortOpt
         .map(_.foldLeft(Seq.empty[Sort])((sorts, value) => sorts :+ SimpleSort(value._1, SortOrder.fromString(value._2))))
         .getOrElse(Nil)
-      new QueryRoot(query, fromOpt, sizeOpt, sorts, timeout, sourceFilter)
+      if (terminateAfter.isDefined) {
```
rcoh:
```
terminateAfter
  .map(tA => new Ex....)
  .getOrElse(new Query Root....)
```

CCheSumo: Updated

Issue #85, opened 2016-11-11
RestlasticSearchClient ActorSystem creation by default causing issues


mvaal:
We are running into an issue right now where we are only able to run a single RestlasticSearchClient on a machine because the ActorSystem being created is always default and we are unable to pass props. 
`implicit val system: ActorSystem = ActorSystem()`
Normally I would just override the val for this and set my own value, but unfortunately with how Scala works, it is still calling the initialization and creating the ActorSystem prior to my value being set (so 2 actor systems are created).
Is there a work around for this?

Otherwise please consider making it configurable.
Possible solutions:
Lazy Init:
`implicit lazy val system: ActorSystem = ActorSystem()`
or implicit parameter pass in:
`class RestlasticSearchClient(endpointProvider: EndpointProvider, signer: Option[RequestSigner] = None, indexExecutionCtx: ExecutionContext = ExecutionContext.Implicits.global, searchExecutionCtx: ExecutionContext = ExecutionContext.Implicits.global)(implicit val system: ActorSystem = ActorSystem()){}`

Thanks for your time.

rcoh:
Sorry about that. I'll make a change to make it a parameter and push it
this evening. Would also accept a pull request.

mvaal:
I have submitted a pull request.  Thanks for the quick response.

rcoh:
@CCheSumo would you mind cutting a new release when you get a chance?

CCheSumo:
@rcoh I will push a new release out today.

mvaal:
Any updates on this?  Thanks.

CCheSumo:
@SplashAttacks cutting the release now. It should be available in half an hour or so.

CCheSumo:
@SplashAttacks release cut has been made. The new release version is 1.0.28. Cheers ~

mvaal:
Looks good, thanks a bunch!

PR #86, opened 2016-11-11
Configurable ActorSystem in RestlasticSearchClient

mvaal:
Updated RestlasticSearchClient system variable to be implicit parameter to allow actor system configuration.


rcoh:
LGTM! We should have a new release cut pretty soon. Closes #85

PR #92, opened 2016-12-17
Add top hits aggregation and order for terms aggregation

claymccoy:


elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/dsl/AggregationDsl.scala
```diff
@@ -42,6 +42,7 @@ trait AggregationDsl extends DslCommons with QueryDsl {
                               size: Option[Int], shardSize: Option[Int],
                               hint: Option[String] = None,
                               name: Option[String] = None,
+                              order: Option[SortOrder] = None,
```
rcoh: can you make order last so we retain backwards compatibility?

rcoh (changes requested):
Just one small change, thanks for your contribution!

elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/dsl/AggregationDsl.scala
claymccoy: Yeah, I almost did that, but it seemed fitting to have nested aggs at the end. I'll be happy to change it though.

claymccoy:
Changed parameter order and rebased with master to get #91 and test locally.

PR #95, opened 2016-12-19
Refactor Test to remove interdependencies

### LINKED ISSUE #93: some tests depend on other test side effects
It is surprising when your passing test starts to fail when run with all tests in the file.
In RestlasticSearchClientTest the tests share state in ES and are isolated by the use of PhrasePrefixQuery on a field. It is not obvious when first working in the file.
A typical way of isolating test data in a database is to clear the data in some way before each new test run. I tried using OneInstancePerTest which would provide a clean ES for each test, but it caused a few tests to fail. Those tests can't be run on their own either. It appears that they depend the side effects of other tests to pass.
It seems worthwhile to make writing tests as easy as possible with few surprises in order to encourage outside contributors to write tests.

rcoh:
I'd like to make a second pass to refactor the tests into multiple files because it's getting out of hand, however, I think this is a marked improvement. The PR enables OneInstancePerTest as suggested by @claymccoy to create Elasticsearch tests that are isolated from one-another.

@CCheSumo Please review.

Closes #93.

### DIFF
--- a/elasticsearch-core/src/test/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClientTest.scala
+++ b/elasticsearch-core/src/test/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClientTest.scala
@@ -165,7 +165,7 @@ class RestlasticSearchClientTest extends WordSpec with Matchers with ScalaFuture
       whenReady(resFut) { res =>
         res.jsonStr should include("doc3")
         res.jsonStr should include("doc4")
-        res.jsonStr should not include ("doc5")
+        res.jsonStr should not include("doc5")
       }
     }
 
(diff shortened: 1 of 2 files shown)


elasticsearch-core/src/test/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClientTest.scala
```diff
@@ -174,14 +165,22 @@ class RestlasticSearchClientTest extends WordSpec with Matchers with ScalaFuture
       whenReady(resFut) { res =>
         res.jsonStr should include("doc3")
         res.jsonStr should include("doc4")
-        res.jsonStr should not include("doc5")
+        res.jsonStr should not include ("doc5")
```
CCheSumo: I think the tradition is not to leave a space between the matcher keyword and the value. Either way, we should make the code style consistent.

elasticsearch-core/src/test/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClientTest.scala
```diff
@@ -174,14 +165,22 @@ class RestlasticSearchClientTest extends WordSpec with Matchers with ScalaFuture
...
+      }.map { doc =>
+        restClient.index(index, tpe, doc)
+      }
+      whenReady(Future.sequence(docFutures)) { _ =>
+        refresh()
+      }
       val fut = restClient.startScrollRequest(index, tpe, QueryRoot(MatchAll, Some(1)))
       val scrollId = whenReady(fut) { resp =>
-        resp.id should not be('empty)
+        resp.id should not be ('empty)
```
CCheSumo: Same as above.

CCheSumo (commented):
Two minor changes. Otherwise Lgtm

PR #96, opened 2016-12-21
upgrade elasticsearch and repair error handling

claymccoy:
Many users are going to be using a newer version of elasticsearch. These commits add a test for elastic search error handling, then upgrade elasticsearch where the test fails. The final commit repairs the test in a way that should work with any elasticsearch version, though it changes the value of ElasticErrorResponse.error.
It looks like upgrading elasticsearch causes a lot of tests to fail, but my intention isn't to upgrade elasticsearch. Though I am using this library in a service with the newer version and I would like to see that. I'm just trying to demonstrate a problem with error handling using newer versions and a potential fix.

### DIFF
--- a/elasticsearch-core/pom.xml
+++ b/elasticsearch-core/pom.xml
@@ -68,23 +68,9 @@
     </dependency>
 
     <dependency>
-      <groupId>org.apache.lucene</groupId>
-      <artifactId>lucene-core</artifactId>
-      <version>4.10.4</version>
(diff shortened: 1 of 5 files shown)


rcoh (commented):
Looks good. I'm going to establish a 2.x branch and publicize it in the Readme.

claymccoy:
This PR was more to demonstrate the problem. I have removed the commit that updated ES in a different PR (#97).

PR #97, opened 2016-12-21
Error handling without es update

claymccoy:
This is similar to #96, but without the ES update. The new error handling will work with multiple versions of ES. This is handy since users of this library are probably using a newer version of ES. It would be nice if this could go in before the next release.


rcoh:
Nice! LGTM

PR #99, opened 2017-01-13
Aws request signer does not set port in Host header correctly

james-turner:
The aws request signer omits the port component of the HttpRequest if it has been supplied and the Host header in the HttpRequest is missing. Without this specified it means you cannot utilise elasticsearch running on AWS under a port other than 80/443, for example if you have a proxy -> aws es, where the proxy operates on a non-80/443 port.


rcoh:
LGTM, thanks for the fix!

rcoh:
We'll have a new version published to maven in the next couple of days

PR #102, opened 2017-02-07
Issue 101.  Fixed issue with multiple sort values.

mvaal:
Sort is now optional and maps a single sort key to a list of Sort values.
Unit test to check multi sort is working properly.

### DIFF
--- a/elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/dsl/QueryDsl.scala
+++ b/elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/dsl/QueryDsl.scala
@@ -35,7 +35,7 @@ trait QueryDsl extends DslCommons with SortDsl {
   case class QueryRoot(query: Query,
                        fromOpt: Option[Int],
                        sizeOpt: Option[Int],
-                       sort: Seq[Sort],
+                       sortOpt: Option[Seq[Sort]],
                        timeout: Option[Int],
                        sourceFilter: Option[Seq[String]])
     extends QueryRootLike {
@@ -53,7 +53,7 @@ trait QueryDsl extends DslCommons with SortDsl {
         fromOpt.map(_from -> _) ++
         sizeOpt.map(_size -> _) ++
         timeout.map(t => _timeout -> s"${t}ms") ++
-        sort.map(_sort -> _.toJson) ++
+        sortOpt.map(_sort -> _.map(_.toJson)) ++
         sourceFilter.map(_source -> _)
     }
   }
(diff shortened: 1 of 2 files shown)


elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/dsl/QueryDsl.scala
```diff
@@ -35,7 +35,7 @@ trait QueryDsl extends DslCommons with SortDsl {
   case class QueryRoot(query: Query,
                        fromOpt: Option[Int],
                        sizeOpt: Option[Int],
-                       sort: Seq[Sort],
+                       sortOpt: Option[Seq[Sort]],
```
rcoh: We need to make this change in a way that doesn't break backwards compatibility (so the arg needs to stay the same).

Can you keep the arg the same and just change toJson to work correctly? I think the fix is `sort.toOption.map(...)` 

`toOption` will make it None if it's empty and Some(list) if it isn't.

rcoh (changes requested):
Thanks for the excellent test! Please see comment below about a fix to maintain backwards compatibility. Thanks for your contribution!

elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/dsl/QueryDsl.scala
mvaal: Sure.  This was actually changed from Option to non-option in revision 9f258d3 (8/22/2016) so I thought it would be OK to switch it back.
If you still want me to change it back, how do I update the pull request?  Sorry, I don't do this often.

rcoh: Ah I see -- I forgot that we had that wrapper layer for backwards compat. Unfortunately, while making this change the first time was backwards compatible, your change isn't backwards compatible (as demonstrated by the fact you had to change the tests). I hope to rework this in 2.0.

To push another change, I think you just need to push it to SplashAttacks:master and it should show up here.

mvaal: Here is the option if you want it to stay the same:
`(if(sort.isEmpty) None else Some(sort)).map(_sort -> _.map(_.toJson))`
I feel it is better suited as an Option though since all the other fields are Options (including sourceFilter which is also a Seq).  Let me know your thoughts, will change to your opinion.

mvaal: Ok, I'll have it out soon, thanks.

rcoh: I agree having it as an option would be better, but backwards compatibility is key -- will change in 2.0

rcoh: I believe you can to sort.toOption to remove the if statement as well.

mvaal:
codecov/patch failed with  0% of diff hit (target 61.15%).  Is this wrong or do I need to modify something?

rcoh:
I think that's a bug. Merging

mvaal:
Thanks!  Will you be able to cut a new release with this change?

CCheSumo:
@SplashAttacks sure. let me create a branch cut with the fix

CCheSumo:
@SplashAttacks release 1.0.29 now contains this fix

mvaal:
Thanks guys, you are always super quick to respond, appreciate it.

PR #104, opened 2017-02-15
Moved hard-coded ask timeout to implicit parameter.

mvaal:
Me again (SplashAttacks).  We are running into an issue where multiple large requests (5MB) are timing out  and need to be able to increase the ask timeout to more 30 seconds.  Currently it is hard coded and would like it moved out to implicit with default (like we did with ActorSystem).
Thanks again!


rcoh:
Nice change. @CCheSumo will have another release out soon.

PR #108, opened 2017-03-20
Added support for bulk deletes with test

agearhart:
Please let me know if you have any issues with this code.  :)

### DIFF
--- a/elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClient.scala
+++ b/elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClient.scala
@@ -135,6 +135,11 @@ class RestlasticSearchClient(endpointProvider: EndpointProvider, signer: Option[
     val bulkOperation = Bulk(documents.map(BulkOperation(update, Some(index -> tpe), _)))
     bulkIndex(bulkOperation)
   }
+  
+  def bulkDelete(index: Index, tpe: Type, documents: Seq[Document]): Future[Seq[BulkItem]] = {
+    val bulkOperation = Bulk(documents.map(BulkOperation(delete, Some(index -> tpe), _)))
+    bulkIndex(bulkOperation)
+  }
 
   def putMapping(index: Index, tpe: Type, mapping: Mapping): Future[RawJsonResponse] = {
     implicit val ec = indexExecutionCtx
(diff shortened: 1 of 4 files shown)


elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClient.scala
```diff
@@ -248,6 +253,7 @@ object RestlasticSearchClient {
     case class BulkItem(_index: String, _type: String, _id: String, status: Int, error: Option[String]) {
       def created = status > 200 && status < 299 && !alreadyExists
       def alreadyExists = error.exists(_.contains("DocumentAlreadyExists"))
+      def deleted = status == 200
```
rcoh: This is a confusing name because the same object is returned by the create action. Can you rename this to `success` and have it be >= 200 <= 299?

elasticsearch-core/src/test/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClientTest.scala
```diff
@@ -168,6 +168,52 @@ class RestlasticSearchClientTest extends WordSpec with Matchers with ScalaFuture
         res.jsonStr should not include("doc5")
       }
     }
+    
+    "Support bulk deletion" in {
```
rcoh: Either add deletion to the previous test, or do the setup without checking the result to avoid duplicate tests.

rcoh (changes requested):
Just a couple fixes

elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClient.scala
```diff
@@ -135,6 +135,11 @@ class RestlasticSearchClient(endpointProvider: EndpointProvider, signer: Option[
     val bulkOperation = Bulk(documents.map(BulkOperation(update, Some(index -> tpe), _)))
     bulkIndex(bulkOperation)
   }
+  
+  def bulkDelete(index: Index, tpe: Type, documents: Seq[Document]): Future[Seq[BulkItem]] = {
```
CCheSumo, line 139: Shouldn't we support delete by document ids instead of requiring `documents: Seq[Document]` to be passed in?

agearhart (commented):
I've renamed BulkItem.deleted to BulkItem.success and combined the tests as suggested.  Please let me know if this is what was intended.

maryammzadeh:
This is awesome! I was hoping there would be support for bulk deletes soon! Thanks!

rcoh (approved):

CCheSumo (commented):
@pnzrdrgoon one more comment

elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClient.scala
agearhart: We can totally do that.  I was just keeping with conventions.  I'll prop that change soon.

PR #111, opened 2017-04-18
Swap out spray for Akka http

seanpquig:
Apologies, didn't mean to edit your description - Russell

### DIFF
--- a/elasticsearch-aws/src/main/scala/com/sumologic/elasticsearch/util/AwsRequestSigner.scala
+++ b/elasticsearch-aws/src/main/scala/com/sumologic/elasticsearch/util/AwsRequestSigner.scala
@@ -179,8 +179,7 @@ class AwsRequestSigner(awsCredentialsProvider: AWSCredentialsProvider, region: S
   }
 
   private def hashedPayloadByteArray(httpRequest: HttpRequest): Array[Byte] = {
-    val dataFuture = AkkaHttpUtil.entityToString(httpRequest.entity)
-    val data = Await.result(dataFuture, 1.second)
+    val data = AkkaHttpUtil.entityToString(httpRequest.entity)
     hashSha256(data)
   }
 
--- a/elasticsearch-aws/src/main/scala/com/sumologic/elasticsearch/util/AwsRequestSigner.scala
+++ b/elasticsearch-aws/src/main/scala/com/sumologic/elasticsearch/util/AwsRequestSigner.scala
@@ -178,7 +178,7 @@ class AwsRequestSigner(awsCredentialsProvider: AWSCredentialsProvider, region: S
   private def hashedPayloadByteArray(httpRequest: HttpRequest): Array[Byte] = {
     implicit val system = ActorSystem()
     implicit val materializer = ActorMaterializer()
-    val dataFuture = AkkaHttpResponseUtil.entityToString(httpRequest.entity)(materializer = materializer)
+    val dataFuture = AkkaHttpUtil.entityToString(httpRequest.entity)(materializer = materializer)
     val data = Await.result(dataFuture, 1.second)
     hashSha256(data)
   }
--- a/elasticsearch-aws/src/main/scala/com/sumologic/elasticsearch/util/AwsRequestSigner.scala
+++ b/elasticsearch-aws/src/main/scala/com/sumologic/elasticsearch/util/AwsRequestSigner.scala
@@ -25,11 +25,16 @@ import java.util.{Calendar, TimeZone}
 import javax.crypto.Mac
 import javax.crypto.spec.SecretKeySpec
 
+import scala.concurrent.Await
+import scala.concurrent.duration._
+
+import akka.actor.ActorSystem
+import akka.http.scaladsl.model.HttpRequest
+import akka.http.scaladsl.model.headers.{Host, RawHeader}
+import akka.stream.ActorMaterializer
 import com.amazonaws.auth.{AWSCredentials, AWSCredentialsProvider, AWSSessionCredentials}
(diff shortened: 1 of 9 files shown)


rcoh:
Thanks for this long needed change! It's not building on CI for some reason -- I assume it builds for you locally?

seanpquig:
It does build, and all tests run for me locally @rcoh.  However, I'm using Java 8.  I can revert to Akka versions that still support Java 7?  Unless there are plans to migrate to Java 8 for the project?

elasticsearch-aws/src/main/scala/com/sumologic/elasticsearch/util/AwsRequestSigner.scala
```diff
@@ -171,7 +179,8 @@ class AwsRequestSigner(awsCredentialsProvider: AWSCredentialsProvider, region: S
   }
 
   private def hashedPayloadByteArray(httpRequest: HttpRequest): Array[Byte] = {
-    val data = httpRequest.entity.asString
+    val dataFuture = AkkaHttpUtil.entityToString(httpRequest.entity)
+    val data = Await.result(dataFuture, 1.second)
```
rcoh: what precisely is being waited for here? Are we waiting for the request to finish or is it just for some sort of string conversion?

elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClient.scala
```diff
@@ -225,16 +226,18 @@ class RestlasticSearchClient(endpointProvider: EndpointProvider, signer: Option[
...
-    val responseFuture: Future[HttpResponse] = (IO(Http) ? request).mapTo[HttpResponse]
+    val responseFuture: Future[HttpResponse] = Http().singleRequest(request)
 
-    responseFuture.map { response =>
+    responseFuture.flatMap { response =>
       logger.debug(f"Got Es response: $response")
+      val entityFuture = AkkaHttpUtil.entityToString(response.entity)
       if (response.status.isFailure) {
-        logger.warn(s"Failure response: ${response.entity.asString.take(500)}")
+        val entityString = Await.result(entityFuture, 100.millis)
```
rcoh: Maybe cleaner to use `.recover` or `.recoverWith` to handle the failure without having to Await on the entity?

elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/util/AkkaHttpUtil.scala
```diff
@@ -0,0 +1,41 @@
...
+    * Helper method for getting string out of Akka HTTP Entity
+    * @param entity HTTP entity
+    * @param streamTimeout timeout for Strict stream completion
+    * @param ec implicit ExecutionContext
+    */
+  def entityToString(entity: ResponseEntity,
+                     streamTimeout: FiniteDuration = 1000.millis)
+                    (implicit ec: ExecutionContext = ExecutionContext.global,
+                     materializer: ActorMaterializer): Future[String] = {
+    entity.toStrict(streamTimeout).map(_.data.utf8String)
```
rcoh: I would prefer doing the `Await` here -- we wait for seemingly arbitrary amounts of time for this in other places. I'd rather standardize in one place even if it makes the API a little less flexible. Does this only return a future as a semantic side effect of akka-http or is there really something that we're waiting for?

rcoh (changes requested):

CCheSumo:
@rcoh Regarding the java version, I have seen two PRs having problem with java 7. As a matter of fact, our infrastructure is also moving to java 8. What do you think about we upgrade the scala/akka version? I will create a PR if that sounds reasonable.

elasticsearch-aws/src/main/scala/com/sumologic/elasticsearch/util/AwsRequestSigner.scala
seanpquig: This is an artifact of all Akka HTTP request/response entities being modeled as streams, so in this instance the full payload string is being pulled from that stream abstraction.  It does not involve an HTTP request/network call.

More on streaming nature of entities: http://doc.akka.io/docs/akka-http/current/scala/http/implications-of-streaming-http-entity.html#implications-of-streaming-http-entities

elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClient.scala
seanpquig: I'll try that 👍

elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/util/AkkaHttpUtil.scala
seanpquig: This future is a side effect of Akka HTTP modeling entities as streams.  I'll try refactor to better consolidate necessary waiting, and think through all the implications.

rcoh:
@CCheSumo upgrading the scala/java version is fine with me

rcoh (approved):

seanpquig:
How do you guys want to coordinate Java 8 switch with Akka versions in this PR?  Java 8 would enable these changes to use stable Akka HTTP and not the old experimental jar.

CCheSumo:
@seanpquig let's keep the changes in this PR. After Java 8 is enabled, we have serval libs we will upgrade, the akka core included.

seanpquig:
Sounds good @CCheSumo.  Thanks.

PR #112, opened 2017-04-18
Update to version 2.0: First Task - Remove ExtendedQueryRoot

CCheSumo:
- Removed ExtendedQueryRoot
- Unified sortOpt in QueryRoot
- Some minor code clean up

This is a breaking change. Needs version 2.0

### DIFF
--- a/elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/dsl/QueryDsl.scala
+++ b/elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/dsl/QueryDsl.scala
@@ -31,12 +31,12 @@ trait QueryDsl extends DslCommons with SortDsl {
   trait Filter extends EsOperation
 
   class QueryRoot(query: Query,
-                       fromOpt: Option[Int],
-                       sizeOpt: Option[Int],
-                       sortOpt: Option[Seq[Sort]],
-                       timeoutOpt: Option[Int],
-                       sourceFilterOpt: Option[Seq[String]],
-                       terminateAfterOpt: Option[Int])
+                  fromOpt: Option[Int],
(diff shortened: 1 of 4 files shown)


rcoh:
Can you fork off a 1.x branch if you haven't already? We need a place to make bug fixes against the 1.0 client until 2.0 is ready.

CCheSumo:
Ah, sure. The fork branch is release-1.0.x-bug-fixes. We can cherry-pick to this branch all the fixes needed.

elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/dsl/QueryDsl.scala
```diff
@@ -63,34 +64,12 @@ trait QueryDsl extends DslCommons with SortDsl {
     def apply(query: Query,
```
rcoh: do we still need apply even when we don't need backwards compat?

rcoh (approved):

elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/dsl/QueryDsl.scala
CCheSumo: This is to supply default of None to optional args. I guess we can do that in the class interface without apply. Updating

rcoh:
@CCheSumo seems like you merged this into master? I thought we had discussed merging into a 2.0 branch..?

CCheSumo:
@rcoh I made a branch for 1.0.x branch and was cherry picking to it. But yeah, making a 2.0 branch and keep main branch at 1.0 makes sense. Reverting this and abandoning the 1.0.x branch.

Issue #77, opened 2016-09-28
Remove default value for "highlight" in ElasticJsonDocument 


CCheSumo:
When we move to version 2.0, where we can introducing breaking changes, default value for "highlight" in ElasticJsonDocument should be removed. Current it is set to None.

CCheSumo:
This has been resolved https://github.com/SumoLogic/elasticsearch-client/pull/115 and will be released with the 2.0 branch

PR #116, opened 2017-04-27
2.0 task2 remove highlight default

CCheSumo:
This will require clients who manually creates a ElasticJosonDocument (most likely for testing) to supply a highlight field or none. Please refer to https://github.com/SumoLogic/elasticsearch-client/issues/77
Here is an example
`val raw = RawSearchResponse(Hits(List(ElasticJsonDocument("index", "type", "id", Some(0.1f), decompose(map).asInstanceOf[JObject], highlight = None)), 1))`


CCheSumo:
plox test

rcoh (approved):

Issue #100, opened 2017-02-01
"search_type" -> "scan" is no longer supported


agearhart:
Attempting to run a startScrollRequest against an AWS Elasticsearch 5.1 instance results in an exception with the message

org.json4s.package$MappingException: No usable value for error
Do not know how to convert JObject(List((root_cause,JArray(List(JObject(List((type,JString(illegal_argument_exception)), (reason,JString(No search type for [scan]))))))), (type,JString(illegal_argument_exception)), (reason,JString(No search type for [scan])))) into class java.lang.String

CCheSumo:
search_type=scan and search_type=count are deprecated in Elasticsearch 2.1. A relative thread https://github.com/elastic/elasticsearch-net/issues/1745 is here.

CCheSumo:
Both `search_type=scan` and `search_type=count` has been deprecated.

Document https://www.elastic.co/guide/en/elasticsearch/reference/2.1/breaking_21_search_changes.html

CCheSumo:
https://github.com/SumoLogic/elasticsearch-client/pull/121

agearhart:
Thanks!  I meant to close this after your first reply, sorry.

PR #121, opened 2017-05-18
Release 2.0.0: Task 4 - remove search_type=scan

CCheSumo:
Both search_type=scan and search_type=count has been deprecated. Document https://www.elastic.co/guide/en/elasticsearch/reference/2.1/breaking_21_search_changes.html (issue here https://github.com/SumoLogic/elasticsearch-client/issues/100)

I am removing `search_type=scan` from the client. It turned out to be a little tricky due to the inconsistent behavior of `scroll` with/without `search_type=scan` specified

- with`search_type=scan` specified, `scroll` returns the id to start from, no data is returned
- without `search_type=scan` specified (same with `search_type` with anything else but `scan`), `scoll` returns the id to start from, and data

The awkwardness is documented here https://www.elastic.co/guide/en/elasticsearch/reference/1.5/search-request-scroll.html#scroll-scan

In order to remove `scan`, I have to change the return type of startScrollRequest and logics around it. For the same reason, ScanAndScrollSource is updated.

### DIFF
--- a/elasticsearch-akka/src/main/scala/com/sumologic/elasticsearch/akkahelpers/ScanAndScrollSource.scala
+++ b/elasticsearch-akka/src/main/scala/com/sumologic/elasticsearch/akkahelpers/ScanAndScrollSource.scala
@@ -64,13 +64,7 @@ class ScanAndScrollSource(index: Index, tpe: Type, query: QueryRoot, scrollSourc
     case Event(ActorPublisherMessage.Cancel, _) => stop()
 
     case Event(GotData(nextId, data), FirstScroll) =>
-      if (data.length == 0) {
-        onComplete()
-        stop()
-      } else {
-        consumeIfPossible(nextId, data)
-      }
-
+      consumeIfNotComplete(nextId, data)
     case Event(ScrollFailure(ex), _) =>
       onError(ScrollFailureException("Failed to start the scroll", ex))
       stop(Failure(ex))
@@ -93,12 +87,7 @@ class ScanAndScrollSource(index: Index, tpe: Type, query: QueryRoot, scrollSourc
       stop(Failure(ex))
 
     case Event(GotData(nextId, data), WaitingForDataWithId(currentId)) =>
-      if (data.length == 0) {
-        onComplete()
-        stop()
-      } else {
-        consumeIfPossible(nextId, data)
-      }
+      consumeIfNotComplete(nextId, data)
   }
 
   whenUnhandled {
@@ -106,6 +95,15 @@ class ScanAndScrollSource(index: Index, tpe: Type, query: QueryRoot, scrollSourc
       logger.warn(s"Unhandled event: $otherEvent, $otherData")
       stay()
 
+  }
+
+  private def consumeIfNotComplete(nextId: ScrollId, data: SearchResponse) = {
+    if (data.length == 0) {
+      onComplete()
+      stop()
+    } else {
+      consumeIfPossible(nextId, data)
+    }
   }
 
   private def consumeIfPossible(id: ScrollId, data: SearchResponse) = {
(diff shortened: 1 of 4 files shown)


elasticsearch-akka/src/main/scala/com/sumologic/elasticsearch/akkahelpers/ScanAndScrollSource.scala
```diff
@@ -53,18 +53,23 @@ class ScanAndScrollSource(index: Index, tpe: Type, query: QueryRoot, scrollSourc
...
 
-    case Event(ScrollStarted(scrollId), NotReady) =>
-      requestMore(scrollId)
+    case Event(GotData(nextId, data), FirstScroll) =>
+      if (data.length == 0) {
+        onComplete()
+        stop()
+      } else {
+        consumeIfPossible(nextId, data)
+      }
```
davidcarltonsumo: This is the same as the `GotData` case further down, so I'd vote for combining them.

elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClient.scala
```diff
@@ -175,18 +176,18 @@ class RestlasticSearchClient(endpointProvider: EndpointProvider, signer: Option[
     runEsCommand(query, s"/${index.name}/${tpe.name}/_query", DELETE)
   }
 
-  def startScrollRequest(index: Index, tpe: Type, query: QueryRoot, resultWindow: String = "1m"): Future[ScrollId] = {
+  def startScrollRequest(index: Index, tpe: Type, query: QueryRoot, resultWindowOpt: Option[String] = None, fromOpt: Option[Int] = None, sizeOpt: Option[Int] = None): Future[(ScrollId, SearchResponse)] = {
```
cddude229: I personally consider it awkward to return a tuple.  It produces more easily read consumer code if it returns a case class.  Specifically, `._1` and `._2` have little meaning, but `.scrollID` and `.searchResponse` would have more context.

Some internal teams have adopted never using tuples as return types in a public API

cddude229 (approved):
Changes look fine from my perspective.  One note on a better API, at least in my opinion

elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClient.scala
```diff
@@ -175,18 +176,20 @@ class RestlasticSearchClient(endpointProvider: EndpointProvider, signer: Option[
     runEsCommand(query, s"/${index.name}/${tpe.name}/_query", DELETE)
   }
 
-  def startScrollRequest(index: Index, tpe: Type, query: QueryRoot, resultWindow: String = "1m"): Future[ScrollId] = {
+  // Scroll requests have optimizations that make them faster when the sort order is _doc.
```
davidcarltonsumo, line 179: I don't see a mention of `_doc` in the code here, though?

davidcarltonsumo:
Seems plausible in general, but one suggestion and one question.

elasticsearch-akka/src/main/scala/com/sumologic/elasticsearch/akkahelpers/ScanAndScrollSource.scala
CCheSumo: Combined the logic https://github.com/SumoLogic/elasticsearch-client/pull/121/commits/b90a1ad670f7738a8e85c4750f7a8482a05738df. I did not completely removed the case as they are at different state of the state machine.

elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClient.scala
CCheSumo: The sort by `_doc` is part of the query. And the client is free to sort by anything basically. `_doc` field is something internal of ES that makes the sorting more efficient. I put another line of comment to call out where to put the sorting. https://github.com/SumoLogic/elasticsearch-client/pull/121/commits/dc6f95ff4b2d8798919efb65d2a667d1d0fa965b

davidcarltonsumo: Hmm.  I guess that's okay given that nothing in the method signature is talking about pre-2.x scanning.

davidcarltonsumo (approved):
lgtm

PR #124, opened 2017-05-18
Release 2.0.0: Task 6 - fix a flaky test

CCheSumo:
Looks like we set the timeout on some tests that causing flakiness. In this PR I remove the timeout in these tests.

### DIFF
--- a/elasticsearch-core/src/test/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClientTest.scala
+++ b/elasticsearch-core/src/test/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClientTest.scala
@@ -111,7 +111,7 @@ class RestlasticSearchClientTest extends WordSpec with Matchers with ScalaFuture
     "Be able to create an index, index a document, and search it" in {
       indexDocs(Seq(Document("doc1", Map("text" -> "here"))))
       val resFut = restClient.query(index, tpe, new QueryRoot(TermQuery("text", "here"),
-        sortOpt = Some(Seq(SimpleSort("_timestamp", DescSortOrder))), timeoutOpt = Some(10)))
+        sortOpt = Some(Seq(SimpleSort("_timestamp", DescSortOrder))), timeoutOpt = None))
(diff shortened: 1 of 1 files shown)


cddude229 (approved):

davidcarltonsumo:
I'm a little nervous about removing it entirely - can we set it to a minute or even 10 minutes instead?  I don't want jobs to just get stuck...

CCheSumo:
Time out set to 5 seconds. That is more than enough to get the test stable. Previous, it was 10ms :(.

elasticsearch-core/src/test/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClientTest.scala
```diff
@@ -111,7 +111,7 @@ class RestlasticSearchClientTest extends WordSpec with Matchers with ScalaFuture
     "Be able to create an index, index a document, and search it" in {
       indexDocs(Seq(Document("doc1", Map("text" -> "here"))))
       val resFut = restClient.query(index, tpe, new QueryRoot(TermQuery("text", "here"),
-        sortOpt = Some(Seq(SimpleSort("_timestamp", DescSortOrder))), timeoutOpt = None))
+        sortOpt = Some(Seq(SimpleSort("_timestamp", DescSortOrder))), timeoutOpt = Some(50000)))
```
davidcarltonsumo: this one is 50 seconds, did you mean 5 seconds here too?

davidcarltonsumo:
happy to go with 5 seconds and see how it works. (Though our Sumo Jenkins is not the only place where I've needed to use a timeout of a minute, so I won't be shocked if we need to set it to longer than 5 seconds.)

davidcarltonsumo (approved):
lgtm except for the presumably unintended 50 seconds on one

davidcarltonsumo:
(not that I have anything against 50 seconds as a value!)

CCheSumo:
Meant to be 5 seconds. Updating

PR #131, opened 2017-07-10
update to snapshot version 2.0.1-SNAPSHOT after release

CCheSumo:
Ah, the versions in master are not good. That's why this PR seems weird. Merging this PR should fix it.


CCheSumo:
Merge it as current master is broken with some wrong version numbers.

PR #132, opened 2017-07-11
Add doucmentExistsById api

CCheSumo:
Add an api to efficiently check if a document already exists as described here https://www.elastic.co/guide/en/elasticsearch/guide/current/doc-exists.html

### DIFF
--- a/elasticsearch-akka/pom.xml
+++ b/elasticsearch-akka/pom.xml
@@ -53,7 +53,7 @@
     <dependency>
       <groupId>com.typesafe.akka</groupId>
       <artifactId>akka-stream-experimental_${scala.version.major}</artifactId>
-      <version>2.0.1-SNAPSHOT</version>
+      <version>2.0.1</version>
     </dependency>
 
     <dependency>
--- a/elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClient.scala
+++ b/elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClient.scala
@@ -124,8 +124,7 @@ class RestlasticSearchClient(endpointProvider: EndpointProvider, signer: Option[
     implicit val ec = indexExecutionCtx
     runEsCommand(NoOp, s"/${index.name}/${tpe.name}/$id", DELETE).map(_.mappedTo[DeleteResponse])
   }
-
-  // Returns true if document exists, returns ElasticErrorResponse with status = 404 otherwise
+  
   def documentExistsById(index: Index, tpe: Type, id: String): Future[Boolean] = {
     implicit val ec = indexExecutionCtx
     runEsCommand(NoOp, s"/${index.name}/${tpe.name}/$id", HEAD).map(_ => true).recover {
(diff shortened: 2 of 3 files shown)


elasticsearch-akka/pom.xml
```diff
@@ -53,7 +53,7 @@
     <dependency>
       <groupId>com.typesafe.akka</groupId>
       <artifactId>akka-stream-experimental_${scala.version.major}</artifactId>
-      <version>2.0.1-SNAPSHOT</version>
```
CCheSumo, line 56: Fixing a mistake introduced by replace all 2.0.1 to 2.0.1-SNAPSHOT

elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClient.scala
```diff
@@ -125,6 +125,12 @@ class RestlasticSearchClient(endpointProvider: EndpointProvider, signer: Option[
     runEsCommand(NoOp, s"/${index.name}/${tpe.name}/$id", DELETE).map(_.mappedTo[DeleteResponse])
   }
 
+  // Returns true if document exists, returns ElasticErrorResponse with status = 404 otherwise
```
kumar-avijit: Why not return `false` when the document doesn't exist?

CCheSumo: Updated

davidcarltonsumo: comment needs update, too

elasticsearch-core/src/test/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClientTest.scala
```diff
@@ -928,6 +928,16 @@ class RestlasticSearchClientTest extends WordSpec with Matchers with ScalaFuture
       resFut.futureValue.rawSearchResponse.highlightAsMaps.head should be(Map("text" -> List("here")))
     }
 
+    "support check where a document exists by document id" in {
+      val doc = Document("doc0001", Map("text" -> "here"))
+      val indexFuture = restClient.bulkIndex(index, tpe, Seq(doc))
+      whenReady(indexFuture) { _ => refresh() }
+      // verify doc0001 exists
+      restClient.documentExistsById(index, tpe, "doc0001").futureValue should be(true)
+      // verify doc0002 does not exist
```
davidcarltonsumo: probably don't need these two comments any more, now that we're returning `false` instead of the 404 stuff?

davidcarltonsumo (approved):
lgtm

elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClient.scala
kumar-avijit: lgtm 👍

CCheSumo: removed comments in test as suggested.

PR #134, opened 2017-07-12
Add a size option to ScanAndScrollSource

CCheSumo:
- There was a 5x performance regression in the new client version 2.0.x (with old server version 1.5). 
- We have to make the new client compatible with old server version for the no down-time upgrade. 
- Regression is centered around scan of documents. 
The root cause seems to be that Elasticsearch defaults to a scan size of 10, which is problematic with massive migrations. https://github.com/elastic/elasticsearch/issues/18253
The previous client version may get around with this due to the very efficient `scan` type which has been removed in the new client version. 

In this PR, I added a sizeOpt to ScanAndScrollSource that controls the scan size.

### DIFF
--- a/elasticsearch-akka/src/main/scala/com/sumologic/elasticsearch/akkahelpers/ScanAndScrollSource.scala
+++ b/elasticsearch-akka/src/main/scala/com/sumologic/elasticsearch/akkahelpers/ScanAndScrollSource.scala
@@ -42,7 +42,7 @@ import org.slf4j.LoggerFactory
  * @param scrollSource Raw ES scroll interface
  */
 
-class ScanAndScrollSource(index: Index, tpe: Type, query: QueryRoot, scrollSource: ScrollClient, sizeOpt: Option[Int])
+class ScanAndScrollSource(index: Index, tpe: Type, query: QueryRoot, scrollSource: ScrollClient, sizeOpt: Option[Int] = None)
   extends ActorPublisher[SearchResponse]
   with FSM[ScanState, ScanData] {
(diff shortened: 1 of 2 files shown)


CCheSumo:
plox test

elasticsearch-akka/src/main/scala/com/sumologic/elasticsearch/akkahelpers/ScanAndScrollSource.scala
```diff
@@ -42,7 +42,7 @@ import org.slf4j.LoggerFactory
  * @param scrollSource Raw ES scroll interface
  */
 
-class ScanAndScrollSource(index: Index, tpe: Type, query: QueryRoot, scrollSource: ScrollClient)
+class ScanAndScrollSource(index: Index, tpe: Type, query: QueryRoot, scrollSource: ScrollClient, sizeOpt: Option[Int])
```
davidcarltonsumo: not sure how serious we want to be about compatibility, but maybe add `None` as a default value?

davidcarltonsumo (approved):
lgtm

kumar-avijit:
lgtm 👍

elasticsearch-akka/src/main/scala/com/sumologic/elasticsearch/akkahelpers/ScanAndScrollSource.scala
CCheSumo: Added None as default value.

PR #136, opened 2017-09-13
BulkIndexActor And ScanAndScrollSource to use right execution context

CCheSumo:
Currently BulkIndexerActor & ScanAndScrollSource is using the global execution context. It should use the ingest execution context passed in as defined in

```
class RestlasticSearchClient(endpointProvider: EndpointProvider, signer: Option[RequestSigner] = None,
                             val indexExecutionCtx: ExecutionContext = ExecutionContext.Implicits.global,
                             searchExecutionCtx: ExecutionContext = ExecutionContext.Implicits.global)
```

### DIFF
--- a/elasticsearch-akka/src/main/scala/com/sumologic/elasticsearch/akkahelpers/ScanAndScrollSource.scala
+++ b/elasticsearch-akka/src/main/scala/com/sumologic/elasticsearch/akkahelpers/ScanAndScrollSource.scala
@@ -22,11 +22,13 @@ import akka.actor.FSM.Failure
 import akka.actor.{FSM, Props}
 import akka.pattern.pipe
 import akka.stream.actor.ActorPublisher
-import com.sumologic.elasticsearch.akkahelpers.ScanAndScrollSource.{ScanState, ScanData}
+import com.sumologic.elasticsearch.akkahelpers.ScanAndScrollSource.{ScanData, ScanState}
 import com.sumologic.elasticsearch.restlastic.RestlasticSearchClient.ReturnTypes.{ScrollId, SearchResponse}
 import com.sumologic.elasticsearch.restlastic.ScrollClient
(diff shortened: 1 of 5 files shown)


elasticsearch-akka/src/main/scala/com/sumologic/elasticsearch/akkahelpers/ScanAndScrollSource.scala
```diff
@@ -49,7 +49,7 @@ class ScanAndScrollSource(index: Index, tpe: Type, query: QueryRoot, scrollSourc
   import akka.stream.actor.ActorPublisherMessage
 
   import ScanAndScrollSource._
-  import scala.concurrent.ExecutionContext.Implicits.global
+  implicit val ec = scrollSource.indexExecutionCtx
```
rcoh: not that it really matters to me, but shouldn't this be the search context?

rcoh (approved):
seems fine

elasticsearch-akka/src/main/scala/com/sumologic/elasticsearch/akkahelpers/ScanAndScrollSource.scala
CCheSumo: @rcoh Very interesting Q. At low level, it is performing a search. Though, scan&scroll is not intended to be used in the same thread pool as regular searches. It is a good point to raise though. It makes sense to use the ingest thread pool in our use case, but not in general case. Let me make it accept a parameter for scan and scroll. User can pass in which threadpool makes more sense in their use cases.

CCheSumo: Code changes for the above discussion here https://github.com/SumoLogic/elasticsearch-client/pull/136/commits/2350a42eb65ec7a134018ba1254b1f1b244f45ec

PR #138, opened 2017-09-28
move serialization of json for bulk request from info to debug

CCheSumo:
During the most recent performance benchmarking with BulkIndexActor, I found the sendRequest in fact takes a significant amount of time for data size of 
1. 2000 documents
2. each document approximately 6.5K in terms of pure text data.
The send request takes more than 1 second 

`2017-09-28 02:59:58,325 -0700 INFO  [hostId=nite-metricsstore-5] [module=metricsstore] [localUserName=metricsstore] [logger=elasticsearch.akkahelpers.BulkIndexerActor$] [thread=pool-30-thread-617] Flushed 2000 messages. esTimeMs=2405; sendRequest='1297ms', createRequest='0ms' `

Looking at the logic, most of the code are wrapped around future. The only expensive piece is the json serialization. I moved this line from info to debug


CCheSumo:
@rcoh I am going to merge it since it is a simple log line change and it is blocking my other work. Will make addition updates if requested.

Issue #142, opened 2017-10-06
deleteDocument API doesn't delete all the matching items


cddude229:
Test:
```
    "Support deleting more than 10 docs" in {
      val insertFutures = (0 to 11).map(i => restClient.index(index, tpe, Document(s"doc$i", Map("text7" -> "here7"))))
      val ir = Future.sequence(insertFutures)
      Await.result(ir, 10.seconds)
      refresh()

      val delFut = restClient.deleteDocument(index, tpe, new QueryRoot(MatchAll))
      Await.result(delFut, 10.seconds)
      refresh()

      val count = Await.result(restClient.count(index, tpe, new QueryRoot(MatchAll)), 10.seconds)
      count should be (0)
    }
```

Failure mode:
```
- should Support deleting more than 10 docs *** FAILED ***
  2 was not equal to 0 (RestlasticSearchClientTest.scala:314)
```

cddude229:
Ok, so, I just remembered the `fromOpt` / `sizeOpt` options on `QueryRoot`, so that helps.  However, the API still won't delete _everything_ if it matches more than the maximum value of `sizeOpt` (10k I think)

Issue #141, opened 2017-10-06
deleteDocument API can fail with obscure error


cddude229:
new test:
```
    "Support deleting a doc that doesn't exist" in {
      val delFut = restClient.deleteDocument(index, tpe, new QueryRoot(TermQuery("text7", "here7")))
      Await.result(delFut, 10.seconds) // May not need Await?
    }
```

Failure mode:
```
- should Support deleting a doc that doesn't exist *** FAILED ***
  com.sumologic.elasticsearch.restlastic.RestlasticSearchClient$ReturnTypes$ElasticErrorResponse: ElasticsearchError(status=400): JString({"error":{"root_cause":[{"type":"parse_exception","reason":"Failed to derive xcontent"}],"type":"parse_exception","reason":"Failed to derive xcontent"},"status":400})
  at com.sumologic.elasticsearch.restlastic.RestlasticSearchClient$$anonfun$runRawEsRequest$1.apply(RestlasticSearchClient.scala:254)
  at com.sumologic.elasticsearch.restlastic.RestlasticSearchClient$$anonfun$runRawEsRequest$1.apply(RestlasticSearchClient.scala:249)
  at scala.util.Success$$anonfun$map$1.apply(Try.scala:237)
  at scala.util.Try$.apply(Try.scala:192)
  at scala.util.Success.map(Try.scala:237)
  at scala.concurrent.Future$$anonfun$map$1.apply(Future.scala:235)
  at scala.concurrent.Future$$anonfun$map$1.apply(Future.scala:235)
  at scala.concurrent.impl.CallbackRunnable.run(Promise.scala:32)
  at scala.concurrent.impl.ExecutionContextImpl$AdaptedForkJoinTask.exec(ExecutionContextImpl.scala:121)
  at scala.concurrent.forkjoin.ForkJoinTask.doExec(ForkJoinTask.java:260)
  ...
```

My best guess is it's this bit:
```
      val documents = Await.result(query(index, tpe, deleteQuery, rawJsonStr = false), 10.seconds).rawSearchResponse.hits.hits.map(_._id)
      bulkDelete(index, tpe, documents.map(Document(_, Map()))).map(res => RawJsonResponse(res.toString))
```

Looks like we fetch matching documents (0 matches) and then delete them.  Some googling (https://github.com/elastic/elasticsearch/issues/8595#issuecomment-331390502) tells me that the `Failure to derive xcontent` occurs when the body is empty - so I think that's the issue here.

[...]

PR #143, opened 2017-10-16
MET-5280 deleteDocument API doesn't delete all the matching items

wojciechatsumo:
I believe there is no optimal solution here. The problem is the return type of deleteDocument.  I'm not sure what should we return in such case. One idea is to return an artificial JSON, obtained by joining all responses, that sounds awful. So I've decided to use different signature and create a new function ("deleteDocuments").

There is one downside though - if one is going to delete really many documents the result is going to blow up the memory. To solve this we could turn deleteDocuments into some kind of streaming operation, but I feel that it would be an overkill.

### DIFF
--- a/elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClient.scala
+++ b/elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClient.scala
@@ -190,6 +190,38 @@ class RestlasticSearchClient(endpointProvider: EndpointProvider, signer: Option[
     }
   }
 
+  def deleteDocuments(index: Index, tpe: Type, deleteQuery: QueryRoot, pluginEnabled: Boolean = false): Future[Map[Index, DeleteResponse]] = {
+    def firstScroll(scId: ScrollId) = startScrollRequest(index, tpe, deleteQuery)
+
+    scrollDelete(index, tpe, ScrollId(""), Map.empty[Index, DeleteResponse], firstScroll)
+  }
+
+  private def scrollDelete(index: Index,
+                           tpe: Type,
+                           scrollId: ScrollId,
+                           acc: Map[Index, DeleteResponse],
+                           scrollingFn: (ScrollId) => Future[(ScrollId, SearchResponse)]): Future[Map[Index, DeleteResponse]] = {
+    implicit val ec = indexExecutionCtx
+    scrollingFn(scrollId).flatMap { case (id, response) =>
+      val rawResponse = response.rawSearchResponse
+      val documentIds = rawResponse.hits.hits.map(_._id)
+
+      if (documentIds.isEmpty) {
+        Future(acc)
+      } else {
+        bulkDelete(index, tpe, documentIds.map(Document(_, Map())))
+          .flatMap { items =>
+            scrollDelete(
+              index,
+              tpe,
+              id,
+              acc ++ items.map(item => Dsl.Index(item._id) -> DeleteResponse(item.success)),
+              (scrollId) => scroll(scrollId))
+          }
+      }
+    }
+  }
+
   // Scroll requests have optimizations that make them faster when the sort order is _doc.
   // Put sort by _doc in query as described in the the following document
   // https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-scroll.html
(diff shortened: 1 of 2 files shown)


rcoh:
If we go this route we should deprecate `deleteDocuments`

elasticsearch-core/src/test/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClientTest.scala
```diff
@@ -936,6 +936,25 @@ class RestlasticSearchClientTest extends WordSpec with Matchers with ScalaFuture
       restClient.documentExistsById(index, tpe, "doc0002").futureValue should be(false)
     }
 
+    "Support deleting more than 10000 docs" in {
+      val insertFutures = (1 to 10011).map(i => restClient.index(index, tpe, Document(s"doc$i", Map("text7" -> "here7"))))
+      val ir = Future.sequence(insertFutures)
+      Await.result(ir, 20.seconds)
+      refresh()
+
+      val delFut = restClient.deleteDocuments(index, tpe, new QueryRoot(MatchAll, sizeOpt = Some(9)))
```
rcoh, line 945: Can you also add a test that shows that the existing `deleteDocument` API does not work?

rcoh (changes requested):
Strategy LGTM. Please add test that demonstrates current (wrong) behavior.

rcoh:
Can you also add some logging if we detect that the old method is going to miss some documents?

davidandrzej:
Thoughts @CCheSumo ?

CCheSumo:
@wojciechatsumo Can you illustrate on why you think the deleteDocument API is not working. The current API tries to support ES server both supporting delete by query https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-delete-by-query.html and ES server that does not support it. 

In the first case, if the delete by query is supported (AWS ES 1.x, and 5.x, but not 2.x: in the 2.x version, the delete query is moved to a plugin and AWS simply did not install that plugin and there is no intension to support it). In this case, we fetch the set of document ids and then do a bulk delete. If there are many docs (millions etc.), it is the client's responsibility to restrict the query to do paging-style delete using "from" and "size" options in the query. Normally, I will recommend the bulk request of size to be 1000 to 5000. 

In the second case, it simply handles any amount of documents to be deleted.

cddude229:
@CCheSumo I believe this is in response to both https://github.com/SumoLogic/elasticsearch-client/issues/141 and https://github.com/SumoLogic/elasticsearch-client/issues/142

cddude229 (approved):
Code changes lgtm.  Whether or not the semantics are correct can be decided by you, Russell, and Chuck.  Thanks for doing this!

wojciechatsumo:
@CCheSumo
When there is no plugin, the old deleteDocument can handle deleting up to 10k documents. As you mentioned - the responsibility of pagination + iteration is forwarded to the user, so therefore I considered that automation of this process and enclosing it behind an API is a purpose of #142. Side effect of dealing with this problem is solving #141.

What I meant before is that unfortunately with current signature the old deleteDocument couldn't easily be improved.

Btw new deleteDocuments is probably doing the same thing as plugin, just on a client side:
https://www.elastic.co/guide/en/elasticsearch/plugins/2.3/plugins-delete-by-query.html

elasticsearch-core/src/test/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClientTest.scala
wojciechatsumo: sure, same test, just replace line 945:
val delFut = restClient.deleteDocuments(index, tpe, new QueryRoot(MatchAll, sizeOpt = Some(9)))
with:
val delFut = restClient.deleteDocument(index, tpe, new QueryRoot(MatchAll, sizeOpt = Some(10011)))

Should fail with an exception: "Result window is too large, from + size must be less than or equal to: [10000] but was [10011]."

But I'm not sure if we want to add such tests to the codebase, since it's assuring feature failure, not success.

CCheSumo:
@wojciechatsumo That is a good point. I was trying to maintain the API lightweight and simple. It looks like the advantages triumphs the disadvantages. Lgtm then

CCheSumo (approved):

elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClient.scala
```diff
@@ -180,12 +180,20 @@ class RestlasticSearchClient(endpointProvider: EndpointProvider, signer: Option[
     runEsCommand(EmptyObject, s"/${index.name}", DELETE)
   }
 
+  @deprecated("Use deleteDocuments instead")
```
rcoh: You need to give a reason clearly explaining the problems with the old version.

rcoh (changes requested):

rcoh (approved):

PR #144, opened 2017-11-20
Support RangeQueries along/as filters

wojciechatsumo:
According to docs this one should be fine.
https://www.elastic.co/guide/en/elasticsearch/reference/2.3/query-dsl-range-filter.html

### DIFF
--- a/elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/dsl/QueryDsl.scala
+++ b/elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/dsl/QueryDsl.scala
@@ -151,7 +151,7 @@ trait QueryDsl extends DslCommons with SortDsl {
     }
   }
 
-  case class RangeQuery(key: String, bounds: RangeBound*) extends Query {
+  case class RangeQuery(key: String, bounds: RangeBound*) extends Query with Filter {
     val _range = "range"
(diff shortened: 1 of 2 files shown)


elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/dsl/QueryDsl.scala
```diff
@@ -151,7 +151,7 @@ trait QueryDsl extends DslCommons with SortDsl {
     }
   }
 
-  case class RangeQuery(key: String, bounds: RangeBound*) extends Query {
+  case class RangeQuery(key: String, bounds: RangeBound*) extends Query with Filter {
```
lchodar: Seems like we can already mix-in this trait in code that uses RangeQuery when instantiating the query. 

Still, I wouldn't mind adding it here by default for convenience.

CCheSumo: A RangeFilter looks good. Although I will create a RangeFilter instead of extending the RangeQuery to be consistent with the existing code style. Also, a Query (with score) and a Filter (no scoring) do have their differences.

CCheSumo (approved):
Lgtm in general. One small comment. Feel free to merge after the comment has been addressed.

PR #145, opened 2017-11-21
Update release & snapshot version after 2.0.5 branch cut

CCheSumo:
Update release & snapshot version after 2.05 branch cut. 

release 2.0.5 has some changes @wojciechatsumo depends on.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,25 +21,25 @@ The library components are offered a la carte:
     <dependency>
       <groupId>com.sumologic.elasticsearch</groupId>
       <artifactId>elasticsearch-core</artifactId>
-      <version>2.0.4</version>
+      <version>2.0.5</version>
     </dependency>
 
     <dependency>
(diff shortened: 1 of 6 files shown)


CCheSumo:
plox test

CCheSumo:
@wojciechatsumo The recently added test is timing out, I think because of travis is using much weaker machines compared with what we have. Shall we reduce the size test? I think as long as we can lock down the scan/scroll delete logic is right, whether it is 10000, or 1000. It does not make a big difference. Need to make sure we do at least two scroll, that is. 

should Support deleting more than 10000 docs *** FAILED ***
  java.util.concurrent.TimeoutException: Futures timed out after [20 seconds]

wojciechatsumo:
Sure, that makes sense take a look at #146, I think I scrolling in "chunks" of 9 was the problem.

PR #148, opened 2017-12-15
Add retry on version conflict option

CCheSumo:
When we have multiple parallel bulkUpdates operations, it can result in version conflict. The default behavior is to throw error. 

In the PR, I added the option of _retry_on_conflict.

### DIFF
--- a/elasticsearch-akka/src/main/scala/com/sumologic/elasticsearch/akkahelpers/BulkIndexerActor.scala
+++ b/elasticsearch-akka/src/main/scala/com/sumologic/elasticsearch/akkahelpers/BulkIndexerActor.scala
@@ -63,8 +63,8 @@ class BulkIndexerActor(restlasticSearchClient: RestlasticSearchClient, bulkConfi
       queue = OpWithTarget(BulkOperation(index, Some(indexName -> tpe), doc), sender(), sess) :: queue
       if (queueFull) flush()
 
-    case UpdateRequest(sess, index, tpe, doc, retryOnVersionConflictOpt) =>
-      queue = OpWithTarget(BulkOperation(update, Some(index -> tpe), doc, retryOnVersionConflictOpt), sender(), sess) :: queue
+    case UpdateRequest(sess, index, tpe, doc, retryOnConflictOpt) =>
+      queue = OpWithTarget(BulkOperation(update, Some(index -> tpe), doc, retryOnConflictOpt), sender(), sess) :: queue
       if (queueFull) flush()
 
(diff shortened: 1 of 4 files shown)


elasticsearch-akka/src/main/scala/com/sumologic/elasticsearch/akkahelpers/BulkIndexerActor.scala
```diff
@@ -63,8 +63,8 @@ class BulkIndexerActor(restlasticSearchClient: RestlasticSearchClient, bulkConfi
       queue = OpWithTarget(BulkOperation(index, Some(indexName -> tpe), doc), sender(), sess) :: queue
       if (queueFull) flush()
 
-    case UpdateRequest(sess, index, tpe, doc) =>
-      queue = OpWithTarget(BulkOperation(update, Some(index -> tpe), doc), sender(), sess) :: queue
+    case UpdateRequest(sess, index, tpe, doc, retryOnVersionConflictOpt) =>
```
kumar-avijit: nit: may be rename this to `retryOnConflictOpt` to be more like the ES parameter?

kumar-avijit (commented):
just a minor comment about renaming. otherwise, lgtm. 👍

elasticsearch-core/src/test/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClientTest.scala
```diff
@@ -213,6 +213,27 @@ class RestlasticSearchClientTest extends WordSpec with Matchers with ScalaFuture
...
+      val docUpdate2 = Document("doc", Map("text" -> "retry on version conflict 2"))
+      val docUpdate3 = Document("doc", Map("text" -> "retry on version conflict 3"))
+
+      val bulkUpdateFuture = (1 to 5).map { _ =>
+        restClient.bulkUpdate(index, tpe, Seq(docUpdate1, docUpdate2, docUpdate3))
+      }
+      Future.sequence(bulkUpdateFuture).futureValue.toString().contains("version conflict") should be(true)
+
+      val bulkUpdateFutureWithRetry = (1 to 5).map { _ =>
+        restClient.bulkUpdate(index, tpe, Seq(docUpdate1, docUpdate2, docUpdate3), retryOnConflictOpt = Some(100))
```
rcoh, line 232: It's really unclear what `Int` means in the context of retryOnConflict. Please add a comment to the method.

PR #151, opened 2018-01-18
Support partition document updates with whole document upserts

CCheSumo:
In the previous implementation of bulkUpdate request, 
```
def bulkUpdate(index: Index, tpe: Type, documents: Seq[Document], retryOnConflictOpt: Option[Int] = None)
```
`doc_as_upsert` is always true. The document we provide in the update request need to be the whole document. 

https://www.elastic.co/guide/en/elasticsearch/reference/2.3/docs-update.html

This PR adds an upsertOpt: Option[Document] = None option to the bulkUpdate request. 
- When upsertOpt = None, defaults to the old behavior
- When upsertOpt = Some(doc_as_upsert), doc provided is used for update/partial update, doc_as_upsert provided is used for index when a document does not already exists.

### DIFF
--- a/elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClient.scala
+++ b/elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClient.scala
@@ -154,6 +154,12 @@ class RestlasticSearchClient(endpointProvider: EndpointProvider, signer: Option[
     val bulkOperation = if(upsertsOpt.isEmpty) {
       Bulk(documents.map(BulkOperation(update, Some(index -> tpe), _, retryOnConflictOpt)))
     } else {
+      val upsertsSize = upsertsOpt.get.size
+      val documentsSize = documents.size
+      if(upsertsSize != documentsSize) {
+        logger.warn(s"Number of documents $documentsSize is different from number of upserts $upsertsSize. " +
+          s"Only first ${Math.min(documentsSize, upsertsSize)} updates in the bulkUpdate request will be applied.")
+      }
       Bulk(documents.zip(upsertsOpt.getOrElse(documents)).map { case (document, upsert) =>
         BulkOperation(update, Some(index -> tpe), document, retryOnConflictOpt, Some(upsert))})
     }
--- a/elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClient.scala
+++ b/elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClient.scala
@@ -148,9 +148,15 @@ class RestlasticSearchClient(endpointProvider: EndpointProvider, signer: Option[
   }
 
   // retryOnConflictOpt specifies how many times to retry before throwing version conflict exception.
-  // https://www.elastic.co/guide/en/elasticsearch/reference/2.3/docs-update.html#_parameters_2
-  def bulkUpdate(index: Index, tpe: Type, documents: Seq[Document], retryOnConflictOpt: Option[Int] = None): Future[Seq[BulkItem]] = {
-    val bulkOperation = Bulk(documents.map(BulkOperation(update, Some(index -> tpe), _, retryOnConflictOpt)))
+  // upsertOpt specifies the document to create in an update request when the document does not exist.
+  // https://www.elastic.co/guide/en/elasticsearch/reference/2.3/docs-update.html
+  def bulkUpdate(index: Index, tpe: Type, documents: Seq[Document], retryOnConflictOpt: Option[Int] = None, upsertsOpt: Option[Seq[Document]] = None): Future[Seq[BulkItem]] = {
+    val bulkOperation = if(upsertsOpt.isEmpty) {
+      Bulk(documents.map(BulkOperation(update, Some(index -> tpe), _, retryOnConflictOpt)))
+    } else {
+      Bulk(documents.zip(upsertsOpt.getOrElse(documents)).map { case (document, upsert) =>
+        BulkOperation(update, Some(index -> tpe), document, retryOnConflictOpt, Some(upsert))})
+    }
     bulkIndex(bulkOperation)
   }
   
--- a/elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClient.scala
+++ b/elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClient.scala
@@ -148,21 +148,9 @@ class RestlasticSearchClient(endpointProvider: EndpointProvider, signer: Option[
   }
 
   // retryOnConflictOpt specifies how many times to retry before throwing version conflict exception.
-  // upsertOpt specifies the document to create in an update request when the document does not exist.
-  // https://www.elastic.co/guide/en/elasticsearch/reference/2.3/docs-update.html
-  def bulkUpdate(index: Index, tpe: Type, documents: Seq[Document], retryOnConflictOpt: Option[Int] = None, upsertsOpt: Option[Seq[Document]] = None): Future[Seq[BulkItem]] = {
-    val bulkOperation = if(upsertsOpt.isEmpty) {
-      Bulk(documents.map(BulkOperation(update, Some(index -> tpe), _, retryOnConflictOpt)))
-    } else {
-      val upsertsSize = upsertsOpt.get.size
-      val documentsSize = documents.size
-      if(upsertsSize != documentsSize) {
-        logger.warn(s"Number of documents $documentsSize is different from number of upserts $upsertsSize. " +
-          s"Only first ${Math.min(documentsSize, upsertsSize)} updates in the bulkUpdate request will be applied.")
-      }
-      Bulk(documents.zip(upsertsOpt.getOrElse(documents)).map { case (document, upsert) =>
-        BulkOperation(update, Some(index -> tpe), document, retryOnConflictOpt, Some(upsert))})
-    }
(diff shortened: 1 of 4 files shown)


CCheSumo:
Looks like @wojciechatsumo is on vacation. @kumar-avijit  do you mind taking a look at this PR?

elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClient.scala
```diff
@@ -148,9 +148,15 @@ class RestlasticSearchClient(endpointProvider: EndpointProvider, signer: Option[
   }
 
   // retryOnConflictOpt specifies how many times to retry before throwing version conflict exception.
-  // https://www.elastic.co/guide/en/elasticsearch/reference/2.3/docs-update.html#_parameters_2
-  def bulkUpdate(index: Index, tpe: Type, documents: Seq[Document], retryOnConflictOpt: Option[Int] = None): Future[Seq[BulkItem]] = {
-    val bulkOperation = Bulk(documents.map(BulkOperation(update, Some(index -> tpe), _, retryOnConflictOpt)))
+  // upsertOpt specifies the document to create in an update request when the document does not exist.
+  // https://www.elastic.co/guide/en/elasticsearch/reference/2.3/docs-update.html
+  def bulkUpdate(index: Index, tpe: Type, documents: Seq[Document], retryOnConflictOpt: Option[Int] = None, upsertsOpt: Option[Seq[Document]] = None): Future[Seq[BulkItem]] = {
```
kumar-avijit: This interface seems fragile because it insists on `documents` being zippable with `upsertsOpt`.

I guess I do not understand why if I want to do an upsert, I have to specify it in 2 places, one in `documents`, and another in `upsertsOpt`?

kumar-avijit: It seems we are trying to 'bulk'-ify this particular api: https://www.elastic.co/guide/en/elasticsearch/reference/2.3/docs-update.html#_literal_doc_as_upsert_literal

Would `bulkUpdate` that, instead of `upsertsOpt`,  takes a boolean flag `isUpsert`, and applies it to all `documents`, work for our use case?
```
def bulkUpdate(index: Index, tpe: Type, documents: Seq[Document], retryOnConflictOpt: Option[Int] = None, isUpsert: Boolean): Future[Seq[BulkItem]]
```

elasticsearch-core/src/test/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClientTest.scala
```diff
@@ -374,6 +374,41 @@ class RestlasticSearchClientTest extends WordSpec with Matchers with ScalaFuture
       }
     }
 
+    "Support bulk partial document updates with whole document upserts" in {
```
kumar-avijit, line 377: Does this test need 4 documents, or can it be shortened to use fewer, may be 2, documents? That would make it easier to understand.

rcoh:
I concur with @kumar-avijit's comments.

elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClient.scala
CCheSumo: Oh, this is the old behavior where the whole document is specified as upsert. As a result, we have to know the entire content of the document. 

The new behavior is trying to support the use case of (partial document update with whole document upsert):
1. We have first_seen_timstamp in the document
2. When we do the update call we want to 
    a. create the document with the timestamp provided if the document does not exist
    b. maintain the old first_seen_timestamp, but update the rest of the document (and we do not know what is the old first_seen_timestamp when we initiate the call)

so the document will be something like "cluster=kafka, metrics=cpu", and the contents in upsert will be "cluster=kafka, metrics=cpu, first_seen_timestamp=124238993"

elasticsearch-core/src/test/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClientTest.scala
CCheSumo: Yeah, good point. Reducing to 2 documents

CCheSumo: test updated here https://github.com/SumoLogic/elasticsearch-client/pull/151/commits/bb0633cae40cbdf5f0866424878f23325c21d391

elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClient.scala
wojciechatsumo: EDIT:
After looking more into the code my thoughts are a bit different, than before.

If we perform operations of the same type across the bulk, then   it should be bulkUpsert() handling upserts, not bulkUpdate().

We went for 'unified' bulk operations exposed to the user, so one can easily create bulk request with operations of the same type.

We got function for handling mixed operations, but it is not very convenient to use:
```
case class Bulk(operations: Seq[BulkOperation])
def bulkIndex(bulk: Bulk): Future[Seq[BulkItem]]
```
because you need to create Bulk, Seq[BulkOperation] and every BulkOperation manually.

wojciechatsumo: Whats more, you could differentiate between:
Upsert with doc_as_upsert -> true
Upsert with upsert-> document_data
and have separate functions for bulking each of them.

CCheSumo:
After offline discussion with @kumar-avijit, we decided to remove the support from bulkIndexAPI as the use case is not a general use case. If the user really need to use partial doc update with whole doc upsert, they can create the BulkOperations themselves and directly call bulkIndex api as demonstrated in the UT. Commit is here https://github.com/SumoLogic/elasticsearch-client/pull/151/commits/18dd05d1e90f4490e2d1660ffcc3e503ecc9fec3

elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/dsl/IndexDsl.scala
```diff
@@ -51,13 +51,20 @@ trait IndexDsl extends DslCommons {
...
     import EsOperation.compactJson
     override def toJson: Map[String, Any] = throw new UnsupportedOperationException
     def toJsonStr: String = {
       val (doc, retryOpt) = operation match {
         case `update` =>
-          (Document(document.id, Map("doc"->document.data) ++ Map("detect_noop" -> true, "doc_as_upsert" -> true)), retryOnConflictOpt.map(n => Map("_retry_on_conflict" -> n)))
+          if (upsertOpt.isEmpty) {
+            (Document(document.id, Map("doc"->document.data) ++ Map("detect_noop" -> true, "doc_as_upsert" -> true)), retryOnConflictOpt.map(n => Map("_retry_on_conflict" -> n)))
+          } else {
+            (Document(document.id, Map("doc"->document.data) ++ Map("detect_noop" -> true) ++ Map("upsert" -> upsertOpt.getOrElse(document).data)), retryOnConflictOpt.map(n => Map("_retry_on_conflict" -> n)))
```
kumar-avijit: nit:
may be replace the need for a `getOrElse` in `"upsert" -> upsertOpt.getOrElse(document).data` by deconstructing `upsertOpt`? So, e.g., replace the `if (upsertOpt.isEmpty)` with a pattern match..

CCheSumo: Good point. Making the change

elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/dsl/IndexDsl.scala
```diff
@@ -51,13 +51,20 @@ trait IndexDsl extends DslCommons {
...
+          }
```
kumar-avijit: Hm, looking at this more carefully, I realize that `BulkOperation` is more generic than I had previously thought, so adding `upsertOpt` as a parameter here doesn't look quite right.

`upsertOpt` makes sense only when the `BulkOperation` is of type `update`, and not otherwise. It seems clunky to me to have an API that allows us to specify that parameter for other kinds of operations.

CCheSumo: That is true from API point of view. Then we will need to add an API for BulkUpdateOperation, which is not too bad I guess.

kumar-avijit: That said, I don't know how to avoid this clunkiness while maintaining backward incompatibility.

One way to fix the issue would have been to have a different case class for each kind of operation, e.g., `BulkUpdateOperation`, `BulkDeleteOperation`, etc.

Doing that, however, would have required changes to `Bulk`s argument types.

kumar-avijit:
Apart from the API clunkiness issue -- certain parameters to `BulkOperation` are ignored when they are not relevant to the `operation` -- this lgtm.

CCheSumo:
@kumar-avijit +1 on the api clunkiness issue. I was trying to break down the api to different type of operations. Cannot make it backward compatible due to prohibited case class inheritance. Going to leave it for now and create an issue to refactor the API for when we upgrade to 6.0.

CCheSumo:
issue https://github.com/SumoLogic/elasticsearch-client/issues/152 created. It can only be executed when breaking changes can be applied.

PR #155, opened 2018-02-01
Update snapshot and version numbers after releaes 2.0.7

CCheSumo:
Update snapshot and version numbers after releaes 2.0.7


lchodar (approved):
LGTM

PR #163, opened 2018-06-28
Support NestedMapping and MultiFieldsMapping in Release 2.0.8.1

CCheSumo:
We haven't moved elastic-client to 6.x in metrics. Plan is to do that next quarter. Meanwhile, we will need to support nested mapping and multi-fields mapping. I am adding this change to this branch only atm. I will create a 2.0.9 branch for publishing after all functionalities has been included.

I am also adding a issue https://github.com/SumoLogic/elasticsearch-client/issues/164 to cherry-pick changes here to the latest master.


lchodar (approved):

wojciechatsumo (approved):
Great! I was about to introduce similar changes, nice that you got it before I even started! :)

PR #166, opened 2018-07-13
Deprecate ScanAndScrollSource and provide ScanAndScrollSourceStage instead

ghost:
Akka 2.5.x has deprecated sources/sinks created from ActorPublisher and ActorSubscriber. This change provides an ScanAndScrollSourceStage which is based off of GraphStage and can be used to create an akka-stream source instead.

[x] Ran `mvn clean compile test`


elasticsearch-akka/src/main/scala/com/sumologic/elasticsearch/akkahelpers/ScanAndScrollSourceStage.scala
```diff
@@ -0,0 +1,87 @@
...
+      private var resultFuture: Future[Either[Throwable, IdAndData]] = _
+
+      override def preStart(): Unit = {
+        resultFuture = scrollSource.startScrollRequest(index, tpe, query, sizeOpt = sizeOpt).map {
+          case (id, data) => Right(IdAndData(id, data))
+        }.recover(recovery)
+      }
+
+      override def onPull(): Unit = {
+        val result = Await.result(resultFuture, timeout)
```
rcoh, line 58: I'm not up to date on semantics, but blocking with Await.result seems dangerous here

elasticsearch-akka/src/main/scala/com/sumologic/elasticsearch/akkahelpers/ScanAndScrollSourceStage.scala
```diff
@@ -1,5 +1,58 @@
...
+  override val shape: SourceShape[SearchResponse] = SourceShape.of(outlet)
+
+  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic =
+    new GraphStageLogic(shape) with OutHandler {
+
+      private var resultFuture: Future[(ReturnTypes.ScrollId, SearchResponse)] = _
+
+      // TODO: Add some failure management that prints log messages, or throws sensible exceptions
+      private def next(): SearchResponse = {
+        val result = Await.result(resultFuture, timeout)
```
CCheSumo: Do we want to do this synchronously?

CCheSumo:
Similar comment on the synchronous await part.

elasticsearch-akka/src/main/scala/com/sumologic/elasticsearch/akkahelpers/ScanAndScrollSourceStage.scala
ghost: I wasn't too sure about this either. But from what I understand, `onPull` will only get called once when the output port is ready to be pushed to. Since this is a source, `onPull` is also the only callback that will be called during normal operation. If we don't push anything at this point, then we'll get stuck. As far as I understand, blocking here doesn't block the rest of the graph the source is connected to.

ghost:
I think I have a workaround for not blocking onPull. Going to close this PR and create a new one with the work around.

PR #167, opened 2018-07-14
Deprecate ScanAndScrollSource and provide ScanAndScrollSourceStage instead

ghost:
Akka 2.5.x has deprecated sources/sinks created from ActorPublisher and ActorSubscriber. This change provides an ScanAndScrollSourceStage which is based off of GraphStage and can be used to create an akka-stream source instead.

### DIFF
--- a/elasticsearch-akka/src/main/scala/com/sumologic/elasticsearch/akkahelpers/ScanAndScrollSourceGraph.scala
+++ b/elasticsearch-akka/src/main/scala/com/sumologic/elasticsearch/akkahelpers/ScanAndScrollSourceGraph.scala
@@ -19,8 +19,9 @@
 package com.sumologic.elasticsearch.akkahelpers
 
 import akka.NotUsed
-import akka.stream.{Graph, SourceShape}
+import akka.stream._
 import akka.stream.scaladsl.{Flow, GraphDSL, Merge, Source, Unzip}
+import akka.stream.stage.{GraphStage, GraphStageLogic, InHandler, OutHandler}
 import com.sumologic.elasticsearch.restlastic.RestlasticSearchClient.ReturnTypes.{ScrollId, SearchResponse}
 import com.sumologic.elasticsearch.restlastic.ScrollClient
 import com.sumologic.elasticsearch.restlastic.dsl.Dsl._
@@ -29,12 +30,36 @@ import scala.concurrent.ExecutionContext
 
 object ScanAndScrollSourceGraph {
 
-  // TODO: Add failure management and shutdown management
+  // TODO: Add failure management
   def apply(index: Index, tpe: Type,
             query: QueryRoot,
             scrollSource: ScrollClient,
             sizeOpt: Option[Int] = None)
            (implicit ec: ExecutionContext): Graph[SourceShape[SearchResponse], NotUsed] = {
+
+    val completionDeciderGraph = new GraphStage[FlowShape[(ScrollId, SearchResponse), (ScrollId, SearchResponse)]] {
+      val in: Inlet[(ScrollId, SearchResponse)] = Inlet[(ScrollId, SearchResponse)]("completionDecider.in")
+      val out: Outlet[(ScrollId, SearchResponse)] = Outlet[(ScrollId, SearchResponse)]("completionDecider.out")
+      override val shape: FlowShape[(ScrollId, SearchResponse), (ScrollId, SearchResponse)] = FlowShape.of(in, out)
+
+      override def createLogic(inheritedAttributes: Attributes): GraphStageLogic =
+        new GraphStageLogic(shape) with InHandler with OutHandler {
+          override def onPull(): Unit = {
+            pull(in)
+          }
+
+          override def onPush(): Unit = {
+            val elem = grab(in)
+            if (elem._2.length == 0) {
+              completeStage()
+            } else {
+              push(out, elem)
+            }
+          }
+
+          setHandlers(in, out, this)
+        }
+    }
 
     GraphDSL.create() { implicit builder: GraphDSL.Builder[NotUsed] =>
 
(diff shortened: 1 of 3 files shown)


elasticsearch-akka/src/main/scala/com/sumologic/elasticsearch/akkahelpers/ScanAndScrollSourceGraph.scala
```diff
@@ -0,0 +1,87 @@
...
+      )
+
+      val merge = builder.add(Merge[(ScrollId, SearchResponse)](2))
+      val unzip = builder.add(Unzip[ScrollId, SearchResponse])
+      val requestMore = builder.add(Flow[ScrollId].mapAsync(5) { id =>
+        scrollSource.scroll(id)
+      })
+      val completionDecider = builder.add(Flow.fromGraph(completionDeciderGraph))
+
+      source ~> merge.in(0)
```
CCheSumo, line 79: This graph is very interesting/tricky. I am okay with it. Wondering if this is how a source implemented in the new syntax though.

CCheSumo: You need  mvn license plugin to fix the license error

elasticsearch-akka/src/main/scala/com/sumologic/elasticsearch/akkahelpers/ScanAndScrollSourceGraph.scala
```diff
@@ -0,0 +1,87 @@
...
+import akka.stream._
+import akka.stream.scaladsl.{Flow, GraphDSL, Merge, Source, Unzip}
+import akka.stream.stage.{GraphStage, GraphStageLogic, InHandler, OutHandler}
+import com.sumologic.elasticsearch.restlastic.RestlasticSearchClient.ReturnTypes.{ScrollId, SearchResponse}
+import com.sumologic.elasticsearch.restlastic.ScrollClient
+import com.sumologic.elasticsearch.restlastic.dsl.Dsl._
+
+import scala.concurrent.ExecutionContext
+
+object ScanAndScrollSourceGraph {
```
kumar-avijit, line 31: This can be done more succinctly, using the Akka-streams `Source` DSL method [`Source.unfoldAsync`](https://doc.akka.io/api/akka/current/akka/stream/scaladsl/Source$.html#unfoldAsync[S,E](s:S)(f:S=%3Escala.concurrent.Future[Option[(S,E)]]):akka.stream.scaladsl.Source[E,akka.NotUsed]) instead of hand-creating a `GraphStage`.

Here's about how you could go about it:

Given are 2 methods from `ScrollClient`, one that starts a scroll, and another that continues the scroll:
(eliding other parameters here...)
```
def start(): Future[(ScrollId, SearchResponse)]
def next(id: ScrollId): Future[(ScrollId, SearchResponse)]
```

Before we can unfold, we need to do 2 things:
1. The source created by`unfoldAsync` terminates when the underlying method returns `None`. Here, we want termination when search response is empty. So,
    ```
    def emptyAsNone[A](resp: (A, SearchResponse)): Option[(A, SearchResponse)] = {
      if (resp._2.length == 0) None
      else Some(resp)
    }
    ```
2. For purposes of passing to `unfoldAsync`, we need to combine the `start` and `next` into a single method, which takes inputs and produces output in terms of `StartOrScrollId` instead of `ScrollId`: 
    ```
    case object Start
    type StartOrScrollId = Either[Start.type, ScrollId]

    def withStartOrScrollId[T](resp: (ScrollId, T)): (StartOrScrollId, T) = {
      (Right(resp._1), resp._2)
    }

    def startOrNext(id: StartOrScrollId)
    : Future[Option[(StartOrScrollId, SearchResponse)]] = {
      id match {
        case Left(_)         => start()       .map (obj => emptyAsNone(withStartOrScrollId(obj)))
        case Right(scrollId) => next(scrollId).map (obj => emptyAsNone(withStartOrScrollId(obj)))
      }
    }
    ```

Finally, all that remains is unfolding to produce a `Source`:
```
val scanAndScrollSource: Source[SearchResponse, Unit] = {
  Source.unfoldAsync[StartOrScrollId, SearchResponse](Left(Start))(startOrNext) 
}
```

ghost: @kumar-avijit whoa this is pretty cool. How do I detect the completion of the stream though? With this we get a endless stream.

kumar-avijit: The `unfold`ing stops when the method being unfolded returns a `None`:
```
def unfoldAsync[S, E](s: S)(f: (S) ⇒ Future[Option[(S, E)]]): Source[E, NotUsed]
```
Btw, I learned today that the Dsl provides a few other variants such as [`unfoldResource`](https://doc.akka.io/api/akka/current/akka/stream/scaladsl/Source$.html#unfoldResource[T,S](create:()=%3ES,read:S=%3EOption[T],close:S=%3EUnit):akka.stream.scaladsl.Source[T,akka.NotUsed]) and [`unfoldResourceAsync`](https://doc.akka.io/api/akka/current/akka/stream/scaladsl/Source$.html#unfoldResourceAsync[T,S](create:()=%3Escala.concurrent.Future[S],read:S=%3Escala.concurrent.Future[Option[T]],close:S=%3Escala.concurrent.Future[akka.Done]):akka.stream.scaladsl.Source[T,akka.NotUsed]) which can come in handy if you want to add pre- and post- stream methods, such as when you want to initialize or close a resource at the start or end of the stream. Just fyi, in case you find yourself needing to do that.

ghost:
@kumar-avijit's workaround is simple enough, uses akka primitives and thus doesn't need to be merged here. I'm going to close this PR.

PR #169, opened 2018-10-10
MET-0 Solve fixed http timeout problem

wojciechatsumo:
in `runRawEsRequest ` we use default timeout when creating aHTTP request to elasticsearch.
Then even though we return a `Future[RawJsonResponse]`, the maximum timeout is fixed to 20s.
User have no way decide to wait longer than 20 s for a whole request to complete.

This PR fixes this problem by passing timeout explicitly.


Ankurkkhurana:
@wojciechatsumo Can you update the description to reflect the PR purpose.

wojciechatsumo:
@Ankurkkhurana Done

PR #171, opened 2018-11-30
Extract common interfaces to support ES 2 and 6.

betenkowski:
This is necessary for a smooth index migration. I'm not sure what the future should look like, we definitely don't want to support all existing ES versions but it has to be possible to migrate from 2 to 6 (internally, we skipped 5 completely) or from 6 to 7 in the future (no need to support 2 at that time).

In short term, `RestlasticSearchClient2` and `RestlasticSearchClient6` should be DRYed.


elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClient2.scala
```diff
@@ -192,7 +192,7 @@ class RestlasticSearchClient2(endpointProvider: EndpointProvider, signer: Option
     implicit val ec = searchExecutionCtx
     val params = Map("scroll" -> resultWindowOpt.getOrElse(defaultResultWindow)) ++ fromOpt.map("from" -> _.toString) ++ sizeOpt.map("size" -> _.toString)
     runEsCommand(query, s"/${index.name}/${tpe.name}/_search", query = UriQuery(params)).map { resp =>
-      val sr = resp.mappedTo[SearchResponseWithScrollId]
+      val sr = resp.mappedTo[SearchResponseWithScrollId]2
```
michaljmatusiak92: `2`?

elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClient2.scala
```diff
@@ -52,15 +52,8 @@ class RestlasticSearchClient2(endpointProvider: EndpointProvider, signer: Option
...
-    // I'm not totally sure why, but you don't specify the type for _suggest queries
-    implicit val ec = searchExecutionCtx
-    val fut = runEsCommand(query, s"/${index.name}/_suggest")
-    fut.map { resp =>
-      val extracted = resp.mappedTo[SuggestResult]
-      extracted.suggestions
-    }
-  }
+  // TODO(bartek, 2018-11-22): Seems unused and doesn't have to be ported now.
+  def suggest(index: Index, tpe: Type, query: SuggestRoot): Future[Map[String,List[String]]] = ???
```
michaljmatusiak92: Isn't it better to throw the exception explicitly?

elasticsearch-core/src/test/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClient6Test.scala
```diff
@@ -26,7 +27,7 @@ class RestlasticSearchClient6Test extends WordSpec with Matchers with BeforeAndA
   override val restClient = RestlasticSearchClient6Test.restClient
 
   "RestlasticSearchClient6" should {
-    behave like restlasticClient(restClient, IndexName, createIndex())
+    behave like restlasticClient(restClient, IndexName, createIndex(), TextType)
```
michaljmatusiak92: How about creating a case class `Settings` and putting there `IndexName` and `TextType` to avoid long parameter list in the argument. Just an idea to improve.

michaljmatusiak92 (approved):

PR #173, opened 2018-12-11
SUMO : Adding IndexedIndex.

Ankurkkhurana:


betenkowski:
The builds have passed on Travis but Github doesn't see it.

PR #176, opened 2019-01-28
Support query profiling

betenkowski:


elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClient.scala
```diff
@@ -68,14 +68,26 @@ trait RequestSigner {
   def withAuthHeader(httpRequest: HttpRequest): HttpRequest
 }
 
-trait RestlasticSearchClient extends ScrollClient {
+abstract class RestlasticSearchClient(searchExecutionCtx: ExecutionContext) extends ScrollClient {
```
wojciechatsumo, line 71: braking changes?

wojciechatsumo (approved):

elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClient.scala
betenkowski: Theoretically yes but I don't think it matters for anyone.

PR #182, opened 2019-02-05
Sumo 103799: refactor deleteByQuery to return number of deleted documents

tgimis:
Refactor deleteByQuery to return number of deleted documents

### DIFF
--- a/elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClient.scala
+++ b/elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClient.scala
@@ -148,7 +148,7 @@ abstract class RestlasticSearchClient(endpointProvider: EndpointProvider, signer
     runEsCommand(NoOp, s"/${index.name}/${tpe.name}/$id", DELETE).map(_.mappedTo[DeleteResponse])
   }
 
-  def deleteByQuery(index: Index, tpe: Type, query: QueryRoot, waitForCompletion: Boolean): Future[RawJsonResponse]
+  def deleteByQuery(index: Index, tpe: Type, query: QueryRoot, waitForCompletion: Boolean): Future[DeleteByQuerySearchResponse]
 
   def documentExistsById(index: Index, tpe: Type, id: String): Future[Boolean] = {
     implicit val ec = indexExecutionCtx
@@ -452,6 +452,8 @@ object RestlasticSearchClient {
     case class IndexAlreadyExistsException(message: String) extends Exception(message)
 
     case class ElasticErrorResponse(error: JValue, status: Int) extends Exception(s"ElasticsearchError(status=$status): ${error.toString}")
+
+    case class DeleteByQuerySearchResponse(deletedDocumentsCount: BigInt)
 
   }
 
--- a/elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClient.scala
+++ b/elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClient.scala
@@ -453,7 +453,7 @@ object RestlasticSearchClient {
 
     case class ElasticErrorResponse(error: JValue, status: Int) extends Exception(s"ElasticsearchError(status=$status): ${error.toString}")
 
-    case class DeleteByQuerySearchResponse(deletedDocumentsCount: BigInt)
+    case class DeleteByQuerySearchResponse(deletedDocumentsCount: Long)
 
   }
 
--- a/elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClient2.scala
+++ b/elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClient2.scala
@@ -37,7 +37,15 @@ class RestlasticSearchClient2(endpointProvider: EndpointProvider, signer: Option
 
   def deleteByQuery(index: Index, tpe: Type, deleteQuery: QueryRoot, waitForCompletion: Boolean): Future[DeleteByQuerySearchResponse] = {
     implicit val ec = indexExecutionCtx
-    deleteDocuments(index, tpe, deleteQuery).map(resp => DeleteByQuerySearchResponse(resp.size))
+
+    deleteDocuments(index, tpe, deleteQuery).map(resp => {
+      val deletedDocumentsCount = if (waitForCompletion) {
+        resp.size
+      } else {
+        -1
+      }
+      DeleteByQuerySearchResponse(deletedDocumentsCount)
+    })
   }
 
   def createIndex(index: Index, settings: Option[IndexSetting] = None): Future[RawJsonResponse] = {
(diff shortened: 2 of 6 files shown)


elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClient.scala
```diff
@@ -453,6 +453,8 @@ object RestlasticSearchClient {
 
     case class ElasticErrorResponse(error: JValue, status: Int) extends Exception(s"ElasticsearchError(status=$status): ${error.toString}")
 
+    case class DeleteByQuerySearchResponse(deletedDocumentsCount: BigInt)
```
michaljmatusiak92: Let's use regular `Long` (any reason why `BigInt`?)

elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClient.scala
```diff
@@ -148,7 +148,7 @@ abstract class RestlasticSearchClient(endpointProvider: EndpointProvider, signer
     runEsCommand(NoOp, s"/${index.name}/${tpe.name}/$id", DELETE).map(_.mappedTo[DeleteResponse])
   }
 
-  def deleteByQuery(index: Index, tpe: Type, query: QueryRoot, waitForCompletion: Boolean): Future[RawJsonResponse]
+  def deleteByQuery(index: Index, tpe: Type, query: QueryRoot, waitForCompletion: Boolean): Future[DeleteByQuerySearchResponse]
```
michaljmatusiak92, line 151: Please check the count in the response in the unit tests

elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClient.scala
tgimis: In JSON response it is BigInt. Maybe lets leave it as in response?

michaljmatusiak92: Or you can use `count.asInstanceOf[BigInt].toLong` I hope

elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClient.scala
tgimis: Done

elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClient.scala
tgimis: Done

elasticsearch-core/src/test/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClientTest.scala
```diff
@@ -1209,7 +1209,8 @@ trait RestlasticSearchClientTest {
       val termQuery = TermQuery("text7", "here7")
 
       val delFut = restClient.deleteByQuery(index, tpe, new QueryRoot(termQuery), true)
```
michaljmatusiak92, line 1210: Please also add a test case with `val delFut = restClient.deleteByQuery(index, tpe, new QueryRoot(termQuery), false)`

In this test case, check only the response, not the count after sending delete request, since it's asynchronous.

tgimis: Done

elasticsearch-core/src/test/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClientTest.scala
```diff
@@ -1206,16 +1206,24 @@ trait RestlasticSearchClientTest {
...
+    "Support deleting by a query" in {
+      val termQuery = TermQuery("text7", "here7")
+      checkDeletingByAQueryResponse(termQuery)
 
       val count1 = Await.result(restClient.count(index, tpe, new QueryRoot(termQuery)), 10.seconds)
       count1 should be(0)
     }
 
+    "Deleting by a query without waiting generates correct response" in {
+      checkDeletingByAQueryResponse(TermQuery("text7", "here7"))
```
michaljmatusiak92: Here in this test we want to test 
`val delFut = restClient.deleteByQuery(index, tpe, new QueryRoot(termQuery), false)`

michaljmatusiak92: I suggest to copy/paste the test case, generally the code duplication is bad, but in the unit tests I think It's better to duplicate the code to make the testing scenario as clear as possible (and make every tests independent).

michaljmatusiak92: And I'm afraid in this case the returned deleted could could be `-1`.

michaljmatusiak92: Which is correct for the asynchronous way.

tgimis: Fixed

elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClient2.scala
```diff
@@ -35,9 +35,9 @@ class RestlasticSearchClient2(endpointProvider: EndpointProvider, signer: Option
 
   override val version = V2
 
-  def deleteByQuery(index: Index, tpe: Type, deleteQuery: QueryRoot, waitForCompletion: Boolean): Future[RawJsonResponse] = {
+  def deleteByQuery(index: Index, tpe: Type, deleteQuery: QueryRoot, waitForCompletion: Boolean): Future[DeleteByQuerySearchResponse] = {
     implicit val ec = indexExecutionCtx
-    deleteDocuments(index, tpe, deleteQuery).map(resp => RawJsonResponse(resp.toString()))
+    deleteDocuments(index, tpe, deleteQuery).map(resp => DeleteByQuerySearchResponse(resp.size))
```
michaljmatusiak92: I think it would be better to return `-1` when `waitForCompletion` is set to `false`. 
```
deleteDocuments(index, tpe, deleteQuery).map(_ => DeleteByQuerySearchResponse(-1))
```
This will make the ES 2.3 and ES 6 tests consistent.

tgimis: Done

michaljmatusiak92 (approved):
🎉

PR #183, opened 2019-02-19
SUMO-104714: Pass a scroll_id in the body instead of an url parameter

michaljmatusiak92:


elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClient.scala
```diff
@@ -146,12 +146,17 @@ abstract class RestlasticSearchClient(searchExecutionCtx: ExecutionContext) exte
...
+                             profile: Boolean = false,
+                             opParams: Map[String, Any] = Map())
                             (implicit ec: ExecutionContext): Future[RawJsonResponse] = {
     val jsonStr = if (profile) {
-      EsOperation.compactJson(op.toJson(version) + ("profile" -> true))
+      EsOperation.compactJson(op.toJson(version) ++ opParams + ("profile" -> true))
     } else {
-      op.toJsonStr(version)
+      if (opParams.nonEmpty) {
+        EsOperation.compactJson(op.toJson(version) ++ opParams)
```
michaljmatusiak92: `EsOperation.compactJson(op.toJson(version) ++ opParams)` throws exceptions for some `runEsCommand` calls (not implemented)

betenkowski (approved):

PR #185, opened 2019-03-21
MET-8230 querying multiple indices

kasia-macioszek:
Querying multiple indices. Same interface as for one index, but with `List[Index]` instead of `Index`.

### DIFF
--- a/elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClient.scala
+++ b/elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClient.scala
@@ -280,6 +280,11 @@ abstract class RestlasticSearchClient(endpointProvider: EndpointProvider, signer
     runEsCommand(EmptyObject, s"/${index.name}/_refresh")
   }
 
+  def refresh(indices: Seq[Index]): Future[RawJsonResponse] = {
+    implicit val ec = indexExecutionCtx
+    runEsCommand(EmptyObject, s"/${indices.map(i => i.name).mkString(",")}}/_refresh")
+  }
+
   def version: EsVersion
 
   protected def runEsCommand(op: RootObject,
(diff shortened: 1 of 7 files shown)


elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClient.scala
```diff
@@ -108,6 +108,20 @@ abstract class RestlasticSearchClient(endpointProvider: EndpointProvider, signer
     }
   }
 
+  def queryIndices(indices: List[Index],
```
michaljmatusiak92: Can we write more generally `Seq[Index]`?

elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClient.scala
```diff
@@ -108,6 +108,20 @@ abstract class RestlasticSearchClient(endpointProvider: EndpointProvider, signer
...
+            tpe: Type,
```
michaljmatusiak92: minor: indent

elasticsearch-core/src/test/scala/com/sumologic/elasticsearch/restlastic/ElasticsearchIntegrationTest.scala
```diff
@@ -41,10 +41,11 @@ trait ElasticsearchIntegrationTest extends BeforeAndAfterAll with ScalaFutures {
...
   def restClient: RestlasticSearchClient
 
-  val IndexName = s"$indexPrefix-${Random.nextLong()}"
+  val IndexName = s"$indexPrefix-${math.abs(Random.nextLong())}"
 
-  protected def createIndex(): Index = {
-    val index = dsl.Dsl.Index(IndexName)
+  protected def createIndices(cnt: Int = 1): IndexedSeq[Index] = {
+    (1 to cnt).map(idx => {
+    val index = dsl.Dsl.Index(s"${IndexName}-${idx}")
```
michaljmatusiak92: minor: indent

michaljmatusiak92:
Looking at the use cases in the metricsstore code, it would be good to add support for:
```
def count(indices: Seq[Index], tpe: Type, query: QueryRoot): Future[Int]
def refresh(indices: Seq[Index]): Future[RawJsonResponse]
def deleteByQuery(indices: Seq[Index], tpe: Type, query: QueryRoot, waitForCompletion: Boolean): Future[RawJsonResponse]
```

See: `ElasticsearchQueryExecutor` class from the metricsstore code

michaljmatusiak92:
LGTM, it would be good to implement `deleteByQuery` to release it in the same version

michaljmatusiak92:
lgtm

PR #186, opened 2019-03-27
Support for scan and scroll for multiple indices

michaljmatusiak92:
Ran the tests locally

### DIFF
--- a/elasticsearch-core/src/test/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClientTest.scala
+++ b/elasticsearch-core/src/test/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClientTest.scala
@@ -301,7 +301,7 @@ trait RestlasticSearchClientTest {
       }
 
       val query = new QueryRoot(MatchAll, sortOpt = Some(Seq(SimpleSort("id", AscSortOrder))))
-      val fut = restClient.startScrollRequest(Seq(index1, index), tpe, query, sizeOpt = Some(5))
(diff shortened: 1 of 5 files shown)


elasticsearch-core/src/test/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClientTest.scala
```diff
@@ -278,6 +278,48 @@ trait RestlasticSearchClientTest {
...
+        Document(s"doc-$n", Map("ct" -> "ct", "id" -> n))
+      }.map { doc =>
+        restClient.index(index1, tpe, doc)
+      }
+      whenReady(Future.sequence(docFutures1)) { _ =>
+        refresh()
+      }
+
+      val query = new QueryRoot(MatchAll, sortOpt = Some(Seq(SimpleSort("id", AscSortOrder))))
+      val fut = restClient.startScrollRequestIndices(Seq(index1, index), tpe, query, sizeOpt = Some(5))
```
betenkowski: Why `index` instead of `index0`?

betenkowski (approved):

elasticsearch-core/src/test/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClientTest.scala
michaljmatusiak92: Good catch, for the readability purpose it should be `index0`, however in the beginning be defined ` val index = indices.apply(0)`

PR #188, opened 2019-04-01
Sampler aggregation support

betenkowski:


elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClient.scala
```diff
@@ -141,6 +141,13 @@ abstract class RestlasticSearchClient(endpointProvider: EndpointProvider, signer
     }
   }
 
+  def sampleAggregation(index: Index, tpe: Type, query: AggregationQuery): Future[BucketAggregationResultBody] = {
```
michaljmatusiak92, line 144: Don't we want to support sample aggregations for multiple indices in the short/long term world?

michaljmatusiak92: Btw. we should ensure that the current Magellan implementation works with two indices. (I did the first check, but I will re-ensure)

elasticsearch-core/src/test/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClientTest.scala
```diff
@@ -1399,6 +1404,39 @@ trait RestlasticSearchClientTest {
...
 
     }
+
+    "Support sampler aggregations" in {
+      val index = {
+        val index = dsl.Dsl.Index(s"$IndexName-sampler-one-shard")
+        val indexSetting = IndexSetting(
+          1, 1, Analyzers(AnalyzerArray(Analyzer(Name("not_analyzed"), Keyword)), FilterArray()), 1)
+        val indexFut = restClient.createIndex(index, Some(indexSetting))
+        indexFut.futureValue
```
michaljmatusiak92, line 1414: Can't we wait got the restClient to create the index? (let's say 10 seconds)

michaljmatusiak92: Ok, `futureValue` also make sense

michaljmatusiak92 (approved):

elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClient.scala
betenkowski: Magellan uses a different way to run sampler aggregations and it will work with more indices.

PR #192, opened 2019-05-07
Replace slow HexOf in AWS signer

michaljmatusiak92:
Using https://github.com/szymonm/scala-benchmarking:

Before:
```
Benchmark        Mode  Cnt     Score    Error  Units
HexTest.ingest  thrpt   20  1461.657 ± 34.452  ops/s
```
After:
```
Benchmark        Mode  Cnt       Score       Error  Units
HexTest.ingest  thrpt   20  457745.938 ± 15106.433  ops/s
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,11 @@
 language: java
+dist: trusty
+jdk:
+  - oraclejdk8
+addons:
+  apt:
+    packages:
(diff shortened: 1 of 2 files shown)


betenkowski (approved):

michaljmatusiak92:
plox test

PR #196, opened 2019-07-22
Support for max_result_window index setting

mkolodziejskiSL:
This PR adds support for `max_result_window` index setting. The default setting is 10000 which is compatible with Elasticsearch's default.

The `max_result_window` is first mentioned in ES docs for version 2.1:
https://www.elastic.co/guide/en/elasticsearch/reference/2.1/index-modules.html

I think it's okay to support only 2.1-2.4 from 2.x branch. If you feel it'd be better to support 2.0 as well, I can change the `max_result_window` to be an optional argument.

### DIFF
--- a/elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/dsl/IndexDsl.scala
+++ b/elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/dsl/IndexDsl.scala
@@ -100,7 +100,8 @@ trait IndexDsl extends DslCommons {
                           numberOfReplicas: Int,
                           analyzerMapping: Analysis,
                           refreshInterval: Int = 1,
-                          preload: Seq[String] = Seq.empty[String])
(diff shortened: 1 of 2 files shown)


betenkowski (approved):

elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/dsl/IndexDsl.scala
```diff
@@ -100,21 +100,24 @@ trait IndexDsl extends DslCommons {
                           numberOfReplicas: Int,
                           analyzerMapping: Analysis,
                           refreshInterval: Int = 1,
-                          preload: Seq[String] = Seq.empty[String])
+                          preload: Seq[String] = Seq.empty[String],
+                          maxResultWindow: Int = 10000)
```
michaljmatusiak92, line 104: lgtm, 
the another approach is to make `maxResultWindow` - `Option[Int]`, if not set the index with the default value (`10k`) is created

PR #204, opened 2020-01-29
Allow custom content types and bump scala version

mfatyga-sumo:


build.gradle
```diff
@@ -30,7 +30,7 @@ subprojects {
     group = "com.sumologic.elasticsearch"
     description = "Elasticsearch Client for Scala that operates against the REST Endpoint"
 
-    version = '7.1.1-SNAPSHOT'
+    version = '7.1.2-SNAPSHOT'
```
fdaca: I think we usually do this version change during the process of releasing. Is version `7.1.1` already released? If not - I think it should stay unchanged.

fdaca (commented):
Not sure about changing the artifact version.
Rest seems ok.

build.gradle
mfatyga-sumo: Ack, I will revert version to 7.1.1 (I assume this change is simple enough to not require version bump then).

PR #206, opened 2020-03-12
Add backwards compatible method for creating requests with implicit content type.

mfatyga-sumo:


michaljmatusiak92 (approved):

mfatyga-sumo:
Please take another look, it turned out the backward compatible type we need here is plaintext (not json).

michaljmatusiak92 (approved):

mfatyga-sumo:
Actually, we only need plaintext in v2 whereas in v6 we should keep on using json.

PR #209, opened 2020-07-13
ConstantScore query & boost support

betenkowski:
Boost support added to:
* `TermQuery`,
* `WildcardQuery`,
* `PrefixQuery`.

`ConstantScore` query support.

### DIFF
--- a/elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/dsl/QueryDsl.scala
+++ b/elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/dsl/QueryDsl.scala
@@ -32,6 +32,14 @@ trait QueryDsl extends DslCommons with SortDsl {
 
   trait Filter extends EsOperation
 
+  trait ValueBoost {
+    val _boost = "boost"
+    val _value = "value"
+    def value: String
+    def boost: Option[Float]
+    def inner: AnyRef = boost.map(b => Map(_value -> value, _boost -> b)).getOrElse(value)
+  }
+
   class QueryRoot(query: Query,
                   fromOpt: Option[Int] = None,
                   sizeOpt: Option[Int] = None,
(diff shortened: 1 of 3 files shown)


elasticsearch-core/src/test/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClient6Test.scala
```diff
@@ -104,8 +104,7 @@ class RestlasticSearchClient6Test extends WordSpec with Matchers with BeforeAndA
       implicit val formats = org.json4s.DefaultFormats
       val docsCount = 10011
       val documents = (1 to docsCount).map(i => Document(s"doc$i", Map("text7" -> "here7")))
-      val bulkInsertResult = restClient.bulkIndex(index, tpe, documents)
-      Await.result(bulkInsertResult, 20.seconds)
+      restClient.bulkIndex(index, tpe, documents).futureValue
```
michaljmatusiak92, line 107: Is `20` seconds not enough?

elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/dsl/QueryDsl.scala
```diff
@@ -220,12 +228,11 @@ trait QueryDsl extends DslCommons with SortDsl {
     override def toJson(version: EsVersion): Map[String, Any] = Map(_lte -> value)
   }
 
-  case class WildcardQuery(key: String, value: String, boost: Option[Float] = None) extends Query {
+  case class WildcardQuery(key: String, value: String, boost: Option[Float] = None) extends Query with ValueBoost {
```
michaljmatusiak92, line 228: Just to be sure: no `override val value/val value` is needed?

michaljmatusiak92 (approved):

elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/dsl/QueryDsl.scala
betenkowski: No, Scala is smart enough to figure this out.

elasticsearch-core/src/test/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClient6Test.scala
betenkowski: It was more about consistency with the rest of the tests than about 20 seconds (and yes, I wasn't determined enough to fix this everywhere).

PR #220, opened 2021-02-02
Allow Queries in place of Filters

mfiglus-sumo:
Changing signatures of some case classes to allow Query usage there and making Filter extend Query to retain backwards compatibility.
Also modified test code a bit to do the cleanup after each test instead of before/after all since the last test wouldn't pass locally due to the number of indices already created on the way to it.

### DIFF
--- a/elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/dsl/QueryDsl.scala
+++ b/elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/dsl/QueryDsl.scala
@@ -30,7 +30,7 @@ trait QueryDsl extends DslCommons with SortDsl {
 
   trait CompoundQuery extends Query
 
-  trait Filter extends EsOperation
+  trait Filter extends Query
 
   trait ValueBoost {
     val _boost = "boost"
@@ -72,7 +72,7 @@ trait QueryDsl extends DslCommons with SortDsl {
     }
   }
 
-  case class FilteredContext(filter: List[Filter]) extends Query {
+  case class FilteredContext(filter: List[Query]) extends Query {
     val _filter = "filter"
     val _query = "query"
     val _searchType = "search-type"
@@ -84,7 +84,7 @@ trait QueryDsl extends DslCommons with SortDsl {
     }
   }
 
-  case class MultiTermFilterContext(filter: Filter*) extends Query {
+  case class MultiTermFilterContext(filter: Query*) extends Query {
     val _filter = "filter"
     val _query = "query"
     val _must = "must"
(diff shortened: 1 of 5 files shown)


elasticsearch-core/src/test/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClientTest.scala
```diff
@@ -836,6 +836,46 @@ trait RestlasticSearchClientTest {
...
+        Document(s"queryFilterDoc-$n", Map("f1" -> n, "f2" -> "field2", "f3" -> "field3"))
+      }.map { doc =>
+        restClient.index(index, tpe, doc)
+      }
+
+      val range = Future.sequence(futures)
+      whenReady(range) { _ =>
+        refresh()
+      }
+      refresh()
```
michaljmatusiak92, line 850: is this extra `refresh()` needed?

elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/dsl/QueryDsl.scala
```diff
@@ -30,7 +30,7 @@ trait QueryDsl extends DslCommons with SortDsl {
 
   trait CompoundQuery extends Query
 
-  trait Filter extends EsOperation
+  trait Filter extends Query
```
michaljmatusiak92, line 33: I'm wondering if `trait FilterQuery extends Query` wound be more in sync with the current naming, but it requires more changes in our repository that uses ES client

michaljmatusiak92 (approved):

elasticsearch-core/src/test/scala/com/sumologic/elasticsearch/restlastic/RestlasticSearchClientTest.scala
mfiglus-sumo: Maybe it isn't, I just copied this part from another test case

elasticsearch-core/src/main/scala/com/sumologic/elasticsearch/restlastic/dsl/QueryDsl.scala
mfiglus-sumo: I wanted to avoid renaming / removing existing traits, classes, etc. to not introduce any breaking changes

michaljmatusiak92: So let's leave the `Filter` name

